PR #254, opened 2014-10-14
Added read timeout to socket

foeken:
After some production runs the code seems to hold up properly. To create a full-proof consumer you will need to set the `recover_from_connection_close` options. If one node is removed (graceful shutdown for maintenance) the connections are closed. Without this setting the consumer will not reconnect.

Perhaps it should default to true?

One issue we encountered was that in some cases the socket would block while reconnecting. I have not been able to find a specific reason to explain why this is happening but the result is a consumer that 'hangs'.

The code below adds the `@read_write_timeout` behaviour to reading too, previously it was just applied to writing to the socket. This will ensure that even in these cases the reconnect mechanism will kick-in.

We found no other issues in production.

**Update** The blocking socket was probably caused by stunnel, which we use to enable SSL with client certificates

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -657,6 +657,9 @@ module Bunny
 
           recover_channels
         end
+      rescue HostListDepleted
+        reset_host_index
+        retry
       rescue TCPConnectionFailedForAllHosts, TCPConnectionFailed, AMQ::Protocol::EmptyResponseError => e
         @logger.warn "TCP connection failed, reconnecting in #{@network_recovery_interval} seconds"
         sleep @network_recovery_interval


jemc:
I have a question about timeouts.  I used `git blame` to check history of commits for the lines in question and it led me here, so I thought I'd comment in this thread rather than raise a new issue.

In the codebase for `Channel`, (almost) all of the calls to wait_on_continuations are surrounded by an enclosing timeout, as in this example:
```ruby
      Bunny::Timeout.timeout(wait_on_continuations_timeout, ClientTimeout) do
        @last_queue_bind_ok = wait_on_continuations
      end
```

Note that `wait_on_continuations_timeout` is just an alias of an alias:
```ruby
    def wait_on_continuations_timeout
      @connection.transport_write_timeout
    end
```

However, in `wait_on_continuations`, we see another timeout for the blocking call:
```ruby
@continuations.poll(@connection.continuation_timeout)
``` 

I'd like to merge these two timeouts into one if possible, and use the "native" timeout for `ContinuationQueue` instead of using a watchdog-thread-based timeout (`Bunny::Timeout`).  One approach would be to determine which timeout of the two is lesser at runtime, then use that one.  However, I'm not sure whether we actually need both timeout values.

It seems most logical to me to just use `@connection.continuation_timeout` and ignore `@connection.transport_write_timeout`, because this is a timeout around a protocol operation, not a write operation.  However, I'm not fully understanding all of @foeken's logic in these commits and this thread, and I didn't want to undo work I didn't understand the reason for.

Is there any reason not to eliminate the extra timeouts here?

michaelklishin:
@jemc your idea can work quite well. Note that continuation timeouts are actually configured by the users, unlike write timeouts, and assume an exception in the calling thread, not the I/O one (although if we make I/O timeouts trigger a recovery, that may be good enough).

Also note that we have a single-threaded mode which is still supported. Shouldn't change anything for timeouts but continuations implementation may look less than obvious because of that.

PR #278, opened 2015-01-24
Update transport.rb

svenyurgensson:
We have to wait more, when linux resolving connection host when primary DNS fails, in my case I have to set `:connect_timeout` to 50 to get it worked. I guess having DEFAULT_CONNECTION_TIMEOUT = 25 will be enough for general case


michaelklishin:
:+1:

PR #281, opened 2015-01-29
Verify peer by default

michaelklishin:
@foeken


foeken:
:+1: I think this is a great solution.

carlhoerberg:
If the test server's cert is generated from `./spec/tls/cacert.pem` then there's no need to disable `verify_peer` in the tests. If the test server's cert isn't generated from `./spec/tls/cacert.pem` then why bother supplying it? Also I'm in heavy favor of having `verify_peer` on, "secure-by-default". Developers already have the same "problem" with `Net::HTTP` et. al.

michaelklishin:
@carlhoerberg host names do not match. The tests are connecting to localhost but that's not what certificates use.

carlhoerberg:
Ok, then why not generate one that has CN=localhost?

michaelklishin:
@carlhoerberg some day :)

carlhoerberg:
```
openssl req -x509 -newkey rsa:2048 -sha256 -nodes -keyout localhost-key.pem -out localhost-cert.pem -subj "/CN=localhost" -days 3650
```

michaelklishin:
We use tls-gen ;)

PR #293, opened 2015-04-11
updates before_build file path in the readme and specs

bartj3:
the readme and comments in the spec had the incorrect path after the file extension was dropped in c998b68


michaelklishin:
Thank you!

PR #297, opened 2015-04-24
Consistent log output when providing logger

### LINKED ISSUE #296: Unexpected log behaviour due to settting of @logger.progname
Hello,

there was unexpected log output after passing `Rails.logger` to Bunny via ```Bunny.new(logger: Rails.logger)```.

The reason was that Bunny sets `@logger.progname = to_s if @logger.respond_to?(:progname)` [here](  
https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L1112).

Would it be possible to delete [this line](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L1112) so that Bunny doesn't change the behaviour of the logger once it gets [passed in](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L150)?

Even in the case that Bunny defines its [own logger](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L1165) I don't see the reason to reassign the attribute since I don't see how it could have changed since the `@logger` was first set.

If deleting the line is not an option. How are other people dealing with this?

bodymindarts:
The progname attribute of the logger was being reassigned.
This made for unexpected output when clients provided a custom logger.
fixes #296

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -144,10 +144,11 @@ module Bunny
       @user            = self.username_from(opts)
       @pass            = self.password_from(opts)
       @vhost           = self.vhost_from(opts)
-      @logfile         = opts[:log_file] || opts[:logfile] || STDOUT
       @threaded        = opts.fetch(:threaded, true)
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
:+1:

PR #309, opened 2015-05-28
Implement ContinuationQueue more efficiently with lower-level primitives.

jemc:
`ContinuationQueue` can be more efficiently implemented in terms of `Array`, `Mutex`, and `ConditionVariable` than the current implementation in terms of `Queue` and `Timeout`.

Specifically, avoiding `Timeout.timeout` means that we can avoid spawning a new "watchdog" thread for every call to `ContinuationQueue#poll` with a timeout duration given.  Instead, we use a `ConditionVariable` which is a lower-level builtin Ruby primitive that allows for efficient signalling of changes of a shared resource to interested waiting threads guarded by a `Mutex`.

`ConditionVariable` is an oft-forgotten but vitally important primitive for efficient concurrency in Ruby.  This is an appropriate pattern to use any time you have one thread waiting for an action in another thread involving a shared resource, with an optional timeout also builtin to the primitive.  In MRI, this primitive is implemented in C, tied to the C implementation of `Mutex`.  In Rubinius, it is built around `Rubinius::Channel`.

### DIFF
--- a/lib/bunny/concurrent/continuation_queue.rb
+++ b/lib/bunny/concurrent/continuation_queue.rb
@@ -6,36 +6,53 @@ module Bunny
     #
     # @private
     class ContinuationQueue
-      def initialize(*args, &block)
-        @q = ::Queue.new(*args)
+      def initialize
+        @q    = []
+        @lock = ::Mutex.new
+        @cond = ::ConditionVariable.new
       end
 
-      def push(*args)
-        @q.push(*args)
+      def push(item)
+        @lock.synchronize do
+          @q.push(item)
+          @cond.signal
+        end
       end
       alias << push
 
       def pop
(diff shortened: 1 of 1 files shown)


michaelklishin:
I should point out that mutexes and conditional variables is [exactly what Queue uses](https://github.com/ruby/ruby/blob/ruby_2_0_0/lib/thread.rb), so the only improvement here is `Timeout.timeout` elimination in `ContinuationQueue#poll`.

michaelklishin:
Thank you.

jemc:
@michaelklishin - You're right, the inner structure of `Queue` is very similar, this implementation just exposes the inherent timeout ability of `ConditionVariable`, as you mentioned.

Thanks so much for the quick merge!

Issue #307, opened 2015-05-25
Don't use Ruby's Timeout


atombender:
Bunny uses `Timeout`, which is [not safe with I/O](http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/).

We're seeing a lot of framing errors with Bunny (RabbitMQ logs errors about unexpected packets) and general instability, which _could_ be explained by Bunny timing out a network operation and not correctly recovering the connection.

For example: Let's say `Queue#pop` is wrapped in a timeout (as it is in `ContinutationQueue#poll`). The connection does a `basic_get`, but it times out. Now the connection is theoretically still waiting for a response, but the client no longer has this knowledge (state), which was local to whatever was polling.

I don't know if this is true about _that_ case (since there is a set of continutation being maintained) and whether Bunny maintains a sufficiently robust state machine to know what a connection is currently doing. But the use of `Timeout` implies that it's not going to be happy whatever it does; I would argue that if you get a `Timeout::Error`, the socket state is pretty much unknown, and to be you have to tear down the entire socket connection in order to be completely safe.

I'm not familiar enough with Bunny's code to say to what extent this is a problem, but I wanted to raise this issue.

michaelklishin:
It has other problems and not efficient as well  (starts a thread per operation) but I am not aware of any alternative for CRuby.

atombender:
Isn’t the solution to use `IO.select` to wait for data?

michaelklishin:
@atombender we [already do](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/cruby/socket.rb#L52). This is not sufficient in plenty of situations.

We have been aware of `Timeout`'s shortcomings for years but there simply aren't any alternatives on CRuby. On JRuby we use `j.u.c.` where possible.

This whole issue sounds speculative. There are perfectly fine reasons for an operation to time out without implying socket closure, most namely resource-driven timeouts. Please dig deeper before you blame `Timeout` for everything.

michaelklishin:
Also: continuation timeouts are configurable. You can bump them to a very high value and effectively only rely socket timeouts. Socket timeouts can probably still be improved, there was at least 3 significant revisions to `Bunny::Transport`. Specific suggestions and pull requests are welcome, as I'm very busy with RabbitMQ itself these days.

carlhoerberg:
[...]

PR #310, opened 2015-05-29
Use Socket.tcp instead of TCPSocket

carlhoerberg:
Socket.tcp has a connect_timeout parameter, so that Timeout#timeout
isn't required. Socket.tcp with connect_timeout was introduced with 
Ruby 2.0.0 (Ruby 1.9 is EOL so it might be acceptable). 

This is a discussion point rather than a mergeable PR, related to #307.


lib/bunny/cruby/socket.rb
```diff
@@ -6,7 +6,7 @@ module Bunny
   #
   # Heavily inspired by Dalli by Mike Perham.
   # @private
-  class Socket < TCPSocket
+  module Socket
```
michaelklishin: Why is this now a module?

michaelklishin:
I'm fine with requiring Ruby 2.0 from now on. The change looks OK. Please remove Ruby 1.9 from CI and update supported Ruby versions everywhere.

carlhoerberg:
Remove a lot of cruby 1.8/1.9 specific stuff, updated travis, the readme etc.

michaelklishin:
This probably can be Bunny 2.0. Not that we have any major API changes…

Issue #312, opened 2015-06-04
comments say Channel#basic_ack has default `multiple = false`; code differs


wconrad:
The implementation of Channel#basic_ack has two required parameters:

     def basic_ack(delivery_tag, multiple)

However, Channel's comments have examples which will only work if the second parameter, `multiple`, has a default of (I guess) `false`:

     #     ch.basic_ack(delivery_info.delivery_tag)

Which is correct?  The docs, or the code?

michaelklishin:
The code. It's a copy-and-paste artifact. We can add a default value for `multiple` (false). Feel free to submit a PR :)

michaelklishin:
Also note that `Channel#basic_ack` is relatively rarely used (primarily by people familiar with other clients). So it took a while to notice.

wconrad:
I'll make  PR to add the default.

wconrad:
It looks like #basic_ack is broken (it's not doing the to_i on the delivery tag).  I'll cook up a PR for that.

#reject, #ack, and #nack are thin wrappers around their #basic_... counterparts.  They wrap the call to the basic method with guarding_against_stale_delivery_tags, like so:

    def nack(delivery_tag, multiple = false, requeue = false)
      guarding_against_stale_delivery_tags(delivery_tag) do
        basic_nack(delivery_tag.to_i, multiple, requeue)
      end
    end

The basic methods do not use that wrapper:

    def basic_nack(delivery_tag, multiple = false, requeue = false)
      raise_if_no_longer_open!
      @connection.send_frame(AMQ::Protocol::Basic::Nack.encode(@id,
          delivery_tag,
          multiple,
          requeue))

      nil
    end

Should the call to `guarding_against_stale_delivery_tags` move into the basic_ methods, or is it intended that the basic_ methods should be able to operate on stale tags?

michaelklishin:
I'm not against moving `guarding_against_stale_delivery_tags`.

michaelklishin:
They are not broken: they are low level methods and don't do everything they possibly could.

wconrad:
Ah.  Then their rdocs are incorrect.  I'll fix 'em.

I'd also like to add tests for the #basic... methods.  The easy thing to do is to copy/modify the tests for the high level functions.  That leaves the tests simple but damp.  I'm not opposed to "damp" in tests-- they shouldn't be dried with a blowtorch--so long as the duplication won't be a source of pain in the future.  What do you think?

[...]

PR #313, opened 2015-06-04
Sprucing up the Channel#basic_... methods

wconrad:
Changes to Channel#basic_ack, #basic_nak, and #basic_reject:

* Protect against stale delivery tags
* Same defaults as the non-basic methods
* Minor rdoc corrections
* Tests detect more failures

See: #312


michaelklishin:
Thank you!

PR #317, opened 2015-06-23
Use non blocking socket write for publishing messages

jafrog:
Uses `@socket.write_nonblock_fully` by default when publishing messages through `basic_publish`.

As I mentioned [in the mailing list](https://groups.google.com/d/msg/ruby-amqp/OCEMGHfzCX4/FupfJUkEwa4J) I'm not sure if there should be more tests around this change. All existing tests pass with `CI=true`.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -564,7 +564,7 @@ module Bunny
         opts[:mandatory],
         false,
         @connection.frame_max)
-      @connection.send_frameset_without_timeout(frames, self)
+      @connection.send_frameset(frames, self)
 
       self
     end


michaelklishin:
Thank you, I'll get to it later this week.

michaelklishin:
One thing we need to double check is what happens when RabbitMQ stops reading from the socket due to internal frame control or a resource-driven alarm, and client's TCP buffers fill up. We've discovered that most clients do not handle this correctly in rabbitmq/rabbitmq-server#156 but current Bunny implementation does.

jafrog:
Well, looks like it doesn't. I caused "memory resource limit alarm" and it blocks publishers. With non-blocking write Bunny throws `Timeout::Error`, tries to reconnect and gets consequently `Exception in the reader loop: AMQ::Protocol::EmptyResponseError: Empty response received from the server.` and `Bunny::ConnectionClosedError: Trying to send frame through a closed connection.`

I'm not sure this issue can be handled differently when using non-blocking write. Arguably it's up to user to send appropriate timeout if such cases are expected?

michaelklishin:
At the very least default timeout has to be higher.

michaelklishin:
I'm not sure I understand the "it doesn't" part. For exceptions that require retrying (e.g. `EWOULDBLOCK`, `EAGAIN`) we properly wait until it is safe to do so using `IO.select` in `Bunny::Socket#write_nonblock_fully`.

So there's only one related improvement I have in mind: for publishes, we need to serialise
all frames before sending them instead of sending them one by one (in which case one write
can be rejected but others will get through, confusing RabbitMQ server parser). Looking into it.

PR #319, opened 2015-06-25
Add :recovery_attempts option to the connection

jafrog:
`:recovery_attempts` options sets the maximum number of attempts Bunny
would make to recover from network failures before giving up. `:recovery_attempts` is effective only
if `:automatically_recover` option is true.

I'm not sure if I handle the number of recovery attempts here correctly. E.g. if `:recovery_attempts => 1` Bunny would actually try to recover *twice* - once initially and once in the `rescue` block. Should it be this way or does it look like an off-by-one error?

Re tests: I had to resort to mocking even though I see Bunny test suite doesn't do it much. I couldn't come up with a better way to test it in a more "integration" fashion. Any suggestions?

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -673,7 +673,7 @@ module Bunny
 
     # @private
     def should_retry_recovery?
-      @recovery_attempts.nil? || @recovery_attempts > 0
+      @recovery_attempts.nil? || @recovery_attempts > 1
     end
 
     # @private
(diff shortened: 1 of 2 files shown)


michaelklishin:
I think to most users it would seem like an off-by-one error. Mocks in this case are OK. Automatic recovery involves a fair share of manual testing anyway.

jafrog:
Changed it so the option shows the actual number of attempts

michaelklishin:
Thank you. `2.0.0.rc2` is coming to a rubygems near ya tomorrow.

PR #323, opened 2015-07-11
Prefetch

tiredpixel:
The code and documentation related to RabbitMQ `prefetch` `global` appears to be incorrect for recent versions of RabbitMQ, in that it is actually supported (https://www.rabbitmq.com/consumer-prefetch.html). These patches extend the already-present references to `global`, surfacing it through `#prefetch`, keeping the same defaults as before, and updating the documentation to describe the current behaviour.

Although I haven't changed it in these patches, I also wonder about the clarity of the wording with regards to prefetch behaviour:

    Channel Qos (Prefetch Level)
    It is possible to control how many messages at most a consumer will be given (before it acknowledges or rejects previously consumed ones). This setting is per channel and controlled via #prefetch.
    (http://reference.rubybunny.info/Bunny/Channel.html)

Although it is indeed set per-channel as such, until I looked into what was actually happening, I've always interpreted this as meaning that the prefetch limit applies across the whole channel—that is, that setting `prefetch=10` and having multiple consumers will result in those consumers all sharing the same limit. But I see the default `global` Bunny sets is `false`, which actually applies the limit individually for each consumer, rather than `true`, which shares the limit (https://www.rabbitmq.com/consumer-prefetch.html). I haven't updated this wording, though, because I'm not convinced it isn't just me who got confused. :)

Peace,
tiredpixel

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -156,6 +156,8 @@ module Bunny
 
     # @return [Integer] active basic.qos prefetch value
     attr_reader :prefetch_count
+    # @return [Integer] active basic.qos prefetch global mode
+    attr_reader :prefetch_global
 
     DEFAULT_CONTENT_TYPE = "application/octet-stream".freeze
     SHORTSTR_LIMIT = 255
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you, please submit a PR that updates `ChangeLog.md` and credit yourself!

tiredpixel:
@michaelklishin, excellent—thank you! :)

I've submitted patch https://github.com/ruby-amqp/bunny/pull/324 with the documentation requested.

Peace,
tiredpixel

PR #331, opened 2015-08-03
require ruby >= 2.0 in gemspec

### LINKED ISSUE #330: Add a Ruby version check
See #327 and #329. We should fail early and with a sensible error message.

carlhoerberg:
Fixes #330


michaelklishin:
Thank you!

PR #338, opened 2015-08-25
Fix pause for default consumer thread pool  

dhananjaysathe:
The Thread.stop is not an object method and operates in the context of the current thread. You cannot use this to pause and unpause the thread in this way. This commit resolves this by using a flag and check this at the beginning of every loop iteration.


michaelklishin:
Can you please submit the same change against the `1.7.x-stable` branch? Thanks.

PR #340, opened 2015-08-25
Get state of work pool more accurately 

dhananjaysathe:
Add a way to fetch actual state of queue. The running method does not provide you with the state of the actual processing happens and simply indicates if the pool has not been paused or killed

This commit introduces 2 methods that exposes relevant part of the internal queue datastructue without making it public using attr_<methods>. the are namely  a) Backlog b) Busy? both these can be used for cleaner shutdown mechanisms (the channle close causes unclen shutdowns as threads are really still active), and making decisions for dynamic scaling workers using queue "backpressure"

### DIFF
--- a/lib/bunny/consumer_work_pool.rb
+++ b/lib/bunny/consumer_work_pool.rb
@@ -42,6 +42,14 @@ module Bunny
       @running
     end
 
+    def backlog
+      @queue.length
+    end
+
+    def busy?
+      !@queue.empty?
+    end
+
     def shutdown
       @running = false
 


dhananjaysathe:
@michaelklishin there is a fundamental problem in the way channel close method works the https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/channel.rb#L232 works 

The channel connection is closed before the pool actually shuts down. Else the executing contexts in the thread pool end up trying to access the closed channel causing bad errors (its hard to predict at which state this may occur , further complicating it )  

I've got a commit ready if you are up for it ?

michaelklishin:
@dhananjaysathe in the Java client, there's a grace period (consumer shutdown timeout, IIRC) in which consumer methods are allowed to run before the pool is forced to shut down. We should do the same thing — but it may be quite a bit trickier because there is no `j.u.c.` in Ruby.

I'm all for making this change.

dhananjaysathe:
Here is a non blocking way to do this without the sleep , but perhaps it may be an overkill https://gist.github.com/dhananjaysathe/33f710ab03a2c68b5021

michaelklishin:
@dhananjaysathe yup, the famous "self pipe" trick. We can also abuse [Bunny::Concurrent::ContinuationQueue](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/concurrent/continuation_queue.rb) for a count down latch with a timeout. I'd prefer this option since it's a primitive we already use. Thoughts?

PR #343, opened 2015-09-01
default recover_from_connection_close to true and document it

### LINKED ISSUE #342: Default recover_from_connection_close to true
During upgrades of RabbitMQ servers, if `recover_from_connection_close` isn't set to true, client's won't reconnect.

carlhoerberg:
Fixes #342


michaelklishin:
@carlhoerberg please update change logs and the site(if needed)

PR #346, opened 2015-09-07
dont initialize SSL a second time

### LINKED ISSUE #345: wrong argument type Bunny::SSLSocket (expected File)
This is sometimes raised when connecting with TLS.

```
TypeError: wrong argument type Bunny::SSLSocket (expected File)
/app/vendor/bundle/ruby/2.0.0/gems/bunny-2.1.0/lib/bunny/transport.rb:366:in `initialize'
/app/vendor/bundle/ruby/2.0.0/gems/bunny-2.1.0/lib/bunny/transport.rb:366:in `new'
/app/vendor/bundle/ruby/2.0.0/gems/bunny-2.1.0/lib/bunny/transport.rb:366:in `wrap_in_tls_socket'
/app/vendor/bundle/ruby/2.0.0/gems/bunny-2.1.0/lib/bunny/transport.rb:296:in `post_initialize_socket'
/app/vendor/bundle/ruby/2.0.0/gems/bunny-2.1.0/lib/bunny/session.rb:272:in `start'
```

I believe the problem is that [this line](https://github.com/ruby-amqp/bunny/blob/2.1.0/lib/bunny/transport.rb#L296) is called a second time, when `@socket` already is a `Bunny::SSLSocketImpl` instead of a `Bunny::Socket` as it is the first time (only a `Socket` is accepted by `OpenSSL::SSL::SSLSocket#new`).

carlhoerberg:
fixes #345

### DIFF
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -292,7 +292,7 @@ module Bunny
     end
 
     def post_initialize_socket
-      @socket = if uses_tls?
+      @socket = if uses_tls? and !@socket.is_a?(Bunny::SSLSocketImpl)
                   wrap_in_tls_socket(@socket)
                 else
                   @socket


michaelklishin:
@carlhoerberg please submit the same against `2.2.x-stable`?

PR #347, opened 2015-09-07
Issue345

carlhoerberg:


michaelklishin:
The original branch is created off of master. Thanks, I will cherry-pick.

PR #358, opened 2015-10-19
Adds explicit `$host` param to the `before_build` script

jafrog:
Setting a host explicitly helps to avoid the problem when rabbitmqctl
resolves host name as a machine name (e.g. Bob's iMac) but then can't
connect to the cluster using this name. For some reason rabbitmqctl can
successfully connect to the localhost cluster only if the name of the cluster
is "rabbit@localhost" rather than "rabbit@Bob's iMac".

I've had this problem when running `before_build` on several machines. I'm not entirely sure it's not due to the idiosyncrasies of my RabbitMQ configuration so please feel free to ignore!

### DIFF
--- a/bin/ci/before_build
+++ b/bin/ci/before_build
@@ -1,11 +1,12 @@
 #!/usr/bin/env ruby
 
 $nodename = ENV.fetch("RABBITMQ_NODENAME", "rabbit")
+$host     = ENV.fetch("RABBITMQ_HOST", "localhost")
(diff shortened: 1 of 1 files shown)


michaelklishin:
This breaks `RABBITMQ_NODENAME`. The idea is that you should override `RABBITMQ_NODENAME` to be `rabbit@bobsimac` instead of having to set two values.

PR #361, opened 2015-12-25
Add Ruby 2.3.0 to version matrix

michaelklishin:
Submitting as a PR to see if it passes first.


michaelklishin:
Will apply directly to master, also see #369.

PR #366, opened 2016-01-13
Fix incorrect comment in docs

fredrik:
The second example of `basic_reject` rejects the message as per the argument `requeue=false`.


michaelklishin:
Thank you.

PR #368, opened 2016-01-18
A test in Bunny::Queue doesn't specify expectation.

userlocalhost2000:
Hi, I report a trivial PR.

There is no matcher processing in a test suite of 'and default exception handler' which is written in `spec/higher_level_api/integration/basic_consume_spec.rb`.
I think this checks default exception handler, logger.error, is called when some `RuntimeError` is raised at `subscribe` block.

So I added a matcher processing to checks avobe. 
According to this, the test suite checks the behavior that default exception handler is called when `RuntimeError` is occurred.


michaelklishin:
I'm not a fan of mocks but fair enough.

michaelklishin:
Thank you!

PR #369, opened 2016-01-19
Fix the problem that rspec fails on Travis CI

userlocalhost2000:
Hi, I'm trying to address the problem that rspec fails on Travis CI.
This is because the failure of registering RabbitMQ user/permission by `bin/ci/before_build`.

The reason why this problem occurred is `-n` parameter of `rabbitmqctl` which is specified in that script.
Environment variable of `RABBITMQ_NODENAME=rabbit@localhost` is specified in `/etc/rabbitmq/rabbitmq-env.conf` which is generated by Travis.
Therefore, a failure of connection is happened.

To prevent it, I deleted this parameter. Because the most of integrated test suites refer to the server in localhsot.

### DIFF
--- a/bin/ci/before_build
+++ b/bin/ci/before_build
@@ -1,11 +1,10 @@
 #!/usr/bin/env ruby
 
-$nodename = ENV.fetch("RABBITMQ_NODENAME", "rabbit")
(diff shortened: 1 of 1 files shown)


michaelklishin:
This indeed helps. The two failures for this PR have to do with dependency installation issues from rubygems.org. Thank you.

PR #370, opened 2016-01-20
Let's wear a badge of build status!

userlocalhost2000:
May I put on another badge of Travis CI to README?
And I think it is good-looking to collect badges on the top of README.


michaelklishin:
Sorry, I see no point in having more badges.

michaelklishin:
I didn't realise the CI status badge was removed at some point. Thanks, I added it back but not at the top.

PR #374, opened 2016-01-26
Cached bundler

thedrow:


.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: ruby
+sudo: false
```
michaelklishin: This is wrong. `sudo` is used because `rabbitmqctl` is used by `bin/ci/before_script`.

michaelklishin:
Please consider investigating how the build system works before you submit pull requests. `sudo` is required and https://github.com/ruby-amqp/amq-protocol/pull/60 also introduced a change that likely was never tested.

thedrow:
Are you sure that rabbitmqctl always requires sudo? I think that rabbitmq is running under the travis user.

michaelklishin:
Not always but it does on Travis. RabbitMQ is provisioned using a Debian package there:
https://github.com/ruby-amqp/bunny/blob/master/bin/ci/before_build

michaelklishin:
Yeah, a quick glance at the build failure for this PR answers the above question.

PR #375, opened 2016-01-26
Only evaluate strings for debug logging messages if required

thedrow:
Most of the times debug messages are not logged so it's a best practice to avoid evaluating strings that are used in debugging log messages unless they are required.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -1488,7 +1488,7 @@ module Bunny
     #
     # @api plugin
     def recover_from_network_failure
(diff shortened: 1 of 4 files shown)


thedrow:
Should we do this for INFO level log message as well?

michaelklishin:
I'm not convinced there's a whole lot of value in doing this even for debug messages. I certainly don't see any point in doing this for info messages: pretty much everybody I know uses Bunny with info logging enabled.

thedrow:
It's simply slightly faster if you don't use the the logger at debug level.

Issue #376, opened 2016-02-03
Repeated automatic recovery after rabbitmq restart


bkanhoopla:
With automatic recovery turned on and heartbeat set to 30 seconds, we noticed that after rabbitmq was killed and restarted, Bunny 2.2.2 would detect and recover from the network failure, and then repeatedly detect and recover from more "network failures" over and over again every ~60 seconds.

We can reproduce the issue from our ubuntu docker container, talking to rabbitmq 3.5.4 running in another docker container, by running `Bunny::new(automatic_recovery: true, heartbeat: 30, log_level: :debug, host: 'mq').start`, then kill and restart the rabbitmq container:

From Bunny:
```
rails@0321f4fb634f:/opt/rails$ irb
irb(main):001:0> require 'bunny'  
=> true
irb(main):002:0> Bunny::new(automatic_recovery: true, heartbeat: 30, log_level: :debug, host: 'mq').start   
D, [2016-02-03T21:43:40.915077 #615] DEBUG -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Sent protocol preamble
D, [2016-02-03T21:43:40.917025 #615] DEBUG -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Sent connection.start-ok
D, [2016-02-03T21:43:40.917439 #615] DEBUG -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Heartbeat interval negotiation: client = 30, server = 580, result = 30
I, [2016-02-03T21:43:40.917554 #615]  INFO -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Heartbeat interval used (in seconds): 30
D, [2016-02-03T21:43:40.917830 #615] DEBUG -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Sent connection.tune-ok with heartbeat interval = 30, frame_max = 131072, channel_max = 65535
D, [2016-02-03T21:43:40.917938 #615] DEBUG -- #<Bunny::Session:0x7f422466bad0 guest@mq:5672, vhost=/, addresses=[mq:5672]>: Sent connection.open with vhost = /
[...]

PR #377, opened 2016-02-04
Fix repeated automatic recovery from network failures after restarting rabbitmq once

bkanhoopla:
Fix Issue #376, by flagging `reader_loop` to stop before calling `recover_from_network_failure` to start another `reader_loop` thread.


michaelklishin:
@bkanhoopla thank you.

PR #379, opened 2016-02-17
Update comment to reflect code

camelpunch:
It looks like the following commit makes the comment outdated:

https://github.com/ruby-amqp/bunny/commit/d35042a5

That is, we used to send frames one by one, but now pack them all into
one object ready for writing.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -935,7 +935,7 @@ module Bunny
       end
     end
 
-    # Sends multiple frames, one by one. For thread safety this method takes a channel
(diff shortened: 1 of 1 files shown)


camelpunch:
Is the channel sync stuff still needed if it's all sent in one payload?

michaelklishin:
@camelpunch I'm not sure I understand your question. What "channel sync stuff"? You get no guarantees from the OS about when your data will be sent and how it is going to be framed by TCP. Synchronous AMQP 0-9-1 methods expect a response either way.

michaelklishin:
`channel.synchronize` is still necessary because concurrent use of connections is not discouraged.

PR #381, opened 2016-02-25
Consumer worker pool threads bubble up Exception instead of going zombie

seamusabshere:
App developers who want to rescue from Exception should do that inside of their own worker code. This change makes sure that they realize that they are getting Exceptions in the first place.


michaelklishin:
I already explained in https://github.com/gocardless/hutch/issues/195 why this change will not be accepted. And please stop saying that threads with unhandled exceptions do not "go zombie". They simply stop.

Bunny consumers are expected to handle exceptions.

PR #384, opened 2016-03-01
Remove unused ivar in Bunny::Queue

camelpunch:
I'm assuming this isn't poked at from outside somehow. This confused me for a second whilst thinking about design of the iOS client.


michaelklishin:
Thank you!

PR #385, opened 2016-03-01
More unused vars

camelpunch:
I ran a (noisy) static analysis checker and found a couple more candidates for deletion. I got bored of wading through the false results, so there are probably more.


michaelklishin:
:+1:

Issue #359, opened 2015-11-16
NoMethodError after reconnect


eflukx:
Bunny is not reconnecting as expected after network issues. Bunny connection was setup using `:automatically_recover=>true`. After reconnecting the following happens. Parsing error of bogus data? I'm using RabbitMQ v 3.4.4 on Ubuntu as AMQP server. The bunny client is running on Windows 7 x64, MRI v2.2.

```
E, [2015-11-16T20:48:09.238787 #180960] ERROR -- #<Bunny::Session:0x45f4788 user@host.name:5672, vhost=/, addresses=[host.name:5672]>: Got an exception when receiving data: An existing connection was forcibly closed by the remote host. (Errno::ECONNRESET)
W, [2015-11-16T20:48:09.313791 #180960]  WARN -- #<Bunny::Session:0x45f4788 user@host.name:5672, vhost=/, addresses=[host.name:5672]>: Recovering from a network failure...
W, [2015-11-16T20:48:19.346365 #180960]  WARN -- #<Bunny::Session:0x45f4788 user@host.name:5672, vhost=/, addresses=[host.name:5672]>: Retrying connection on next host in line: host.name:5672
W, [2015-11-16T20:48:34.160212 #180960]  WARN -- #<Bunny::Session:0x45f4788 user@host.name:5672, vhost=/, addresses=[host.name:5672]>: TCP connection failed, reconnecting in 5.0 seconds
W, [2015-11-16T20:48:44.200787 #180960]  WARN -- #<Bunny::Session:0x45f4788 user@host.name:5672, vhost=/, addresses=[host.name:5672]>: Retrying connection on next host in line: host.name:5672
C:/Ruby22/lib/ruby/gems/2.2.0/gems/amq-protocol-2.0.0/lib/amq/protocol/frame.rb:62:in `decode_header': undefined method `unpack' for #<Hash:0x45f4260> (NoMethodError)
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/bunny-2.2.1/lib/bunny/transport.rb:245:in `read_next_frame'
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/bunny-2.2.1/lib/bunny/reader_loop.rb:68:in `run_once'
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/bunny-2.2.1/lib/bunny/reader_loop.rb:35:in `block in run_loop'
[...]

PR #387, opened 2016-03-15
Don't recover from network failures in Transport#read_fully

djmb:
Fixes #359 

When recovering from a network failure, `read_fully` returns the return value of the call to `@session.handle_network_failure(e)`.

https://github.com/ruby-amqp/bunny/blob/641088faaf1718b05bad2af5c65bbb925565848e/lib/bunny/transport.rb#L221

In some cases this is a `Hash` returned from here:
https://github.com/ruby-amqp/bunny/blob/641088faaf1718b05bad2af5c65bbb925565848e/lib/bunny/session.rb#L705

It makes sense to avoid recovering from network failures inside `Transport#read_fully` as the recovery code replaces the transport itself. The reader loop can catch the failure and do the recovery for us.

PR #390, opened 2016-03-28
Don't recover from network failures in Transport#read_fully (backport to 1.7.x stable)

kiran:
Fixes https://github.com/ruby-amqp/bunny/issues/359 for 1.7.x-stable.

(cherry-picked commits from https://github.com/ruby-amqp/bunny/pull/387)


michaelklishin:
Any reason to not simply upgrade? Is it an old Ruby version?

PR #398, opened 2016-04-10
fixed ReaderLoop#stopped? and ReaderLoop#stopping?

### LINKED ISSUE #397: Seemingly incorrect implementation of ReaderLoop#stopped?
There is a piece of code that should check status in reader_loop.rb

```
    def stop
      @mutex.synchronize { @stopping = true }
    end

    def stopped?
      @mutex.synchronize { @stopped = true }
    end

    def stopping?
      @mutex.synchronize { @stopping = true }
    end
```

Is this intended that `ReaderLoop#stopped?` updates value of variable `@stopped`?

Zhomart:
Fixes #397

### DIFF
--- a/lib/bunny/reader_loop.rb
+++ b/lib/bunny/reader_loop.rb
@@ -92,11 +92,11 @@ module Bunny
     end
 
     def stopped?
-      @mutex.synchronize { @stopped = true }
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you.

PR #406, opened 2016-05-19
Reset channel unconfirmed_set on network recovery

biinari:
With publisher confirms and automatic recovery enabled, when a network failure occurs, acks / nacks no longer get pushed onto the confirms continuations queue. This causes future `wait_for_confirms` to fail, even though the messages published since network recovery have been acknowledged by the broker.

I have included a spec with this PR that catches the problem and a fix that resets `unconfirmed_set` on network recovery. I tried to keep the stress spec as concise as possible but it does take around 20-30 seconds to run.

Any suggestions to improve this are welcome. In particular I wasn't sure whether to put my spec into `spec/issues` instead.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -1495,6 +1495,7 @@ module Bunny
     # @api plugin
     def recover_confirm_mode
       if using_publisher_confirmations?
+        @unconfirmed_set.clear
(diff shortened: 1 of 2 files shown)


michaelklishin:
Makes sense. Thank you.

biinari:
Thanks @michaelklishin for the quick merge.

PR #411, opened 2016-06-05
Test with Rubinius 3

thedrow:


michaelklishin:
I have failures locally and Travis also fails.

michaelklishin:
I'm afraid I cannot get things to pass on Rubinius 3 :(

PR #414, opened 2016-06-08
Convert socket to Celluloid::IO sockets when required

### LINKED ISSUE #413: Add support for Celluloid::IO sockets
Bunny does not currently expose a mechanism to inject a custom socket class through it's API.
In order to cooperate with Celluloid::IO's reactor we either need that API or to provide support for it by checking if `Celluloid::IO.evented?` is equal to true.
What do you think?

thedrow:
This PR aims to provide support for cooperation with Celluloid::IO's reactor when it is in use.
Fixes #413

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ group :development do
   gem "ruby-prof", :platform => :mri
 
   gem "json",      :platform => :ruby_18
+  gem "celluloid"
   gem "celluloid-io"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ group :development do
   gem "ruby-prof", :platform => :mri
 
   gem "json",      :platform => :ruby_18
+  gem "celluloid-io"
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,9 @@ group :development do
   gem "ruby-prof", :platform => :mri
 
   gem "json",      :platform => :ruby_18
+end
+
+group :ci do
   gem "celluloid"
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -29,6 +29,7 @@ group :development do
   gem "ruby-prof", :platform => :mri
 
   gem "json",      :platform => :ruby_18
+  gem "celluloid-io"
```
michaelklishin: Doesn't this mean that every test run will use `celluloid-io`? Since most users don't use that project, I'd very much prefer to use regular standard library sockets by default in our tests.

thedrow: Sockets won't convert to celluloid-io's sockets unless you're using the reactor.
I'll make the celluloid tests optional. I'm still working on them.

michaelklishin: Thank you!

thedrow:
Ok I added some basic tests. They are not optional yet. How do you do that in RSpec?

michaelklishin:
@thedrow you can use example filtering (tags) or an environment variable. Bunny tends to do the latter for no particular reason.

thedrow:
I'm confused. The tests run fine locally.

michaelklishin:
Note that Travis runs bundler with `--without development`.

michaelklishin:
@thedrow do you think that excluding the `development` group can be problematic here?

michaelklishin:
Yes, the Celluloid spec breaks the rest of the test suite. I think making Celluloid opt-in via options (which are propagated to `Bunny::Socket`) is the right thing to do => closing this iteration.

PR #420, opened 2016-06-15
Allow switching the socket class

### LINKED ISSUE #413: Add support for Celluloid::IO sockets

thedrow:
This is another attempt to fix #413.
It's not clear what we should do about SSLSocket though since Celluloid provides an implementation for it as well. Any ideas?

Also this doesn't have a test just yet but it will.

### DIFF
--- a/lib/bunny/cruby/socket.rb
+++ b/lib/bunny/cruby/socket.rb
@@ -13,12 +13,13 @@ module Bunny
     WRITE_RETRY_EXCEPTION_CLASSES = [Errno::EAGAIN, Errno::EWOULDBLOCK, IO::WaitWritable]
 
     def self.open(host, port, options = {})
-      socket = ::Socket.tcp(host, port, nil, nil,
+      @socket_impl = options.fetch(:socket_impl, ::Socket)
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you. We can either do the same for TLS sockets or switch to supporting specifically Celluloid which will then be used in both places. The latter is OK (I doubt we will have many alternative socket implementations) but the former won't tie us to Celluloid.

thedrow:
Bunny's SSLSocket inherits from OpenSSL's SSLSocket which doesn't allow me to replace the implementation of the socket.

michaelklishin:
@thedrow we can refactor that part and extract an "interface." I'm open to suggestions :) We can make `Bunny::Transport` pluggable and with a base class, too, allowing you to implement any socket implementation logic you please. Note that this will have more significant ramifications for automated testing.

michaelklishin:
No activity on this one and looks like TLS is not covered by this => closing.

PR #422, opened 2016-06-22
Fixes #421. Exchange network reconnect `no_declare` respect

### LINKED ISSUE #421: Exchanges not respecting no_declare on `recover_from_network_failure`
After a network recovery exchange calls `declare!` (https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/exchange.rb#L223), even if the exchange has `opts` set for `no_declare` as is respected upon initialize (https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/exchange.rb#L87).

arfrank:
Adds in respecting `no_declare` in exchange reconnect options.

### DIFF
--- a/lib/bunny/exchange.rb
+++ b/lib/bunny/exchange.rb
@@ -220,7 +220,7 @@ module Bunny
 
     # @private
     def recover_from_network_failure
-      declare! unless predefined?
+      declare! unless opts[:no_declare] || predeclared?
 
       @bindings.each do |b|
         bind(b[:source], b[:opts])
--- a/lib/bunny/exchange.rb
+++ b/lib/bunny/exchange.rb
@@ -220,7 +220,7 @@ module Bunny
 
     # @private
     def recover_from_network_failure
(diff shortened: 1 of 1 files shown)


lib/bunny/exchange.rb
```diff
@@ -220,7 +220,7 @@ def wait_for_confirms
 
     # @private
     def recover_from_network_failure
-      declare! unless predefined?
```
michaelklishin: predeclared or predefined?

michaelklishin:
Please add a test case that at least makes sure the code we modify is executed. A failing case would involve two users, one of which has limited permissions but even a trivial case would catch the typo in the current PR.

lib/bunny/exchange.rb
michaelklishin: Heh, so they are aliases.

michaelklishin:
I'm going to apply this manually using the existing predicate name (`predefined?`) to look less confusing in the diff plus a test in a bit. Thank you for your time!

PR #430, opened 2016-07-19
Make wait_for_confirms poll while there're unconfirmed msgs

### LINKED ISSUE #424: Channel#wait_for_confirms may return before all confirmations for outstanding publishes are received
I discovered a possible race condition in `Channel#wait_for_confirms` which can lead to `wait_for_confirms` returning before RabbitMQ has acknowledged (or nack'ed) the published message. This actually hit us in production.

Suppose the following happens:

1. We have a channel in "confirm select" mode.
2. We do a `basic_publish` to that channel. Some time passes, say >0.5s.
3. During that time, an ack is received by Bunny and handled in `handle_ack_or_nack`. `@unconfirmed_set` is emptied, `true` is pushed to `@confirms_continuations`.
3. We do a second `baslic_publish` to that same channel.
4. Immediately afterwards we issue `wait_for_confirms` to wait for all confirmations. It sees that `@unconfirmed_set` has something in it and issues `@confirms_continuations.poll` but that returns immediately and unexpectedly because of the `true` value pushed to it at step 3. We haven't received an ack or nack for the second publish yet.

The way I understand the documentation is that doing a bunch of publishes and then calling `wait_for_confirms` is a completely valid use case, but unless I'm missing something obvious, it's pretty evident from the code that each `ack` that comes will result in another `true` value pushed to `@confirms_continuations` and that many `wait_for_confirms` will immediately succeed afterwards, no matter how many messages are in `@unconfirmed_set`.

In our code we actually do a publish and `wait_for_confirms` immediately after that, but the race condition can still occur (and it did).

I've also written [a small script that demonstrates the issue](https://gist.github.com/mitio/60448a8e9d4d9180fda63e22ba974b9c).

Tested on: Bunny 2.4.0 and Ruby 2.1.2 on macOS

mitio:
This is an attempt to fix #424.

This should fix a race condition in which Channel#wait_for_confirms may
return before all confirmations are received when using a threaded
connection.

Suppose the following happens:

1. We have a channel in "confirm select" mode.
2. We do a basic_publish to that channel. Some time passes, say a few
   seconds.
3. During that time, an ack is received by Bunny and handled in
   handle_ack_or_nack. @unconfirmed_set is emptied, true is pushed to
   @confirms_continuations.
4. We do a second baslic_publish to that same channel.
5. Immediately afterwards we issue wait_for_confirms to wait for all
   confirmations. It sees that @unconfirmed_set has something in it and
   issues @confirms_continuations.poll but that returns immediately and
   unexpectedly because of the true value pushed to it at step 3. We
   haven't received an ack or nack for the second publish yet.

Doing a bunch of publishes and then calling `wait_for_confirms` is a
completely valid use case and it should block and wait until all acks or
nacks are received. Even if one does a publish and immediately follows
it with `wait_for_confirms`, a race condition exists and may result in
`wait_for_confirms` returning before ack/nack is received.

This is not an issue for non-threaded connections. That is why the spec
includes a test only for threaded connections.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -1790,11 +1790,9 @@ module Bunny
         @threads_waiting_on_confirms_continuations << t
 
         begin
-          outstanding_confirms = false
-          @unconfirmed_set_mutex.synchronize do
-            outstanding_confirms = !@unconfirmed_set.empty?
+          while @unconfirmed_set_mutex.synchronize { !@unconfirmed_set.empty? }
+            @confirms_continuations.poll(@connection.continuation_timeout)
           end
-          @confirms_continuations.poll(@connection.continuation_timeout) if outstanding_confirms
         ensure
           @threads_waiting_on_confirms_continuations.delete(t)
         end
--- a/spec/higher_level_api/integration/publisher_confirms_spec.rb
+++ b/spec/higher_level_api/integration/publisher_confirms_spec.rb
@@ -124,6 +124,59 @@ describe Bunny::Channel do
     end
 
     include_examples "publish confirms"
+
+    it "returns only when all confirmations for publishes are received" do
+      ch = connection.create_channel
+
+      operations_log = []
+      operations_log_mutex = Mutex.new
+      acks_received = Queue.new
+
+      log_acks = proc do |tag, _, is_nack|
+        operations_log_mutex.synchronize do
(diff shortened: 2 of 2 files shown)


michaelklishin:
Thank you, this seems to fix exactly the race condition that has been plaguing our test suite for months: I can run the suite 10 times in a row without failures.

I feel `2.5.0` won't take too long to be released.

PR #437, opened 2016-10-07
Handle graceful shutdown of ConsumerWorkPool when canceling a consumer

stefansedich:
This carries on the work of #341 to implement this feature fully. You can now configure a worker pool shutdown timeout when creating the channel (off by default).

Then upon processing of the basic_cancel instead of killing the worker pool as previously done, the worker pool will be shut down and instructed to wait for any workers to complete (within the timeout period).

This allows for a consumer to stop receiving new messages but still be able to complete any existing work before the worker pool is shut down.

All existing behavior remains in the case of closing a connection, we will not wait for workers as the worker might need to ack the message and in this scenario the connection has been closed and this would not be possible.

I will update the documentation as a seperate PR.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -337,14 +337,14 @@ module Bunny
     # opened (this operation is very fast and inexpensive).
     #
     # @return [Bunny::Channel] Newly opened channel
-    def create_channel(n = nil, consumer_pool_size = 1, consumer_pool_abort_on_exception = false)
+    def create_channel(n = nil, consumer_pool_size = 1, consumer_pool_abort_on_exception = false, consumer_pool_shutdown_timeout = nil)
       raise ArgumentError, "channel number 0 is reserved in the protocol and cannot be used" if 0 == n
 
       @channel_mutex.synchronize do
         if n && (ch = @channels[n])
           ch
         else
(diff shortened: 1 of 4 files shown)


lib/bunny/session.rb
```diff
@@ -337,14 +337,14 @@ def transport_write_timeout
     # opened (this operation is very fast and inexpensive).
     #
     # @return [Bunny::Channel] Newly opened channel
-    def create_channel(n = nil, consumer_pool_size = 1, consumer_pool_abort_on_exception = false)
+    def create_channel(n = nil, consumer_pool_size = 1, consumer_pool_abort_on_exception = false, consumer_pool_shutdown_timeout = nil)
```
michaelklishin: Again, a non-nil timeout (maybe up to 60 seconds?) sounds OK and would make most people benefit from this improvement.

stefansedich: Ok sounds good, I wasn't sure so left it off but I think this makes sense :)

michaelklishin:
It looks good at first glance. I'd recommend using a timeout by default, e.g. the same the Java client uses.

stefansedich:
Changed to use a 60s timeout as mentioned, I wanted to not change behavior but after you mentioned it I think it makes sense to be safe by default.

stefansedich:
@michaelklishin any more thoughts?

michaelklishin:
Thank you.

stefansedich:
Thanks, I will update the documentation when I get a spare minute through the week.

Issue #440, opened 2016-10-27
Support for TLS SNI


c-knowles:
The compose.io implementation of RabbitMQ needs SNI on the TLS connection in order to verify correctly, at least on their very latest implementation. I was looking through the bunny docs and code but didn't spot any support for it. Is there any? If not I may submit the change for review.

The openssl equivalent would be along the lines of adding `servername`:
`openssl s_client -connect HOST:PORT -servername HOST`

It's [undocumented](https://github.com/ruby/openssl/issues/81) but Ruby seems to support this via `hostname` on the SSL socket:

``` ruby
socket = OpenSSL::SSL::SSLSocket.new(tcp_client, ssl_context)
socket.hostname = 'HOST'
socket.connect
```

michaelklishin:
@c-knowles thanks. Do you know what's the earliest Ruby version to provide `OpenSSL::SSL::SSLSocket#hostname`?

michaelklishin:
Three Bunny classes that are relevant if you are to consider a PR:

 * [Bunny::Transport](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/transport.rb#L359)
 * [Bunny::Socket](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/cruby/ssl_socket.rb) for MRI
 * [Bunny::Socket](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/cruby/ssl_socket.rb) for JRuby (JRuby support for Bunny is a "best effort" kind of feature since there's a dedicated JRuby client, March Hare)

Thank you.

c-knowles:
Not sure about the versions yet, I do know it seems to have been in there since at least 2015. So if there is no support yet or planned I will likely have a go at a pull request soon, doesn't look like the changes would be that big.

michaelklishin:
We can always set that field only if the socket object responds to `:hostname`.

c-knowles:
According to https://github.com/ruby/openssl/issues/81 on the `hostname` setting:
```
Yes, it is part of the public API, since Ruby 1.9.2 (note it requires OpenSSL >= 0.9.8f).

RDoc comment was fixed some time ago, and OpenSSL::SSL::SSLSocket#hostname= 
now shows up in the latest HTML generated from the master branch:

http://ruby.github.io/openssl/OpenSSL/SSL/SSLSocket.html#method-i-hostname-3D
```

lockyluchiano:
1

PR #442, opened 2016-10-31
Docker rabbitmq setup works better out of the box

c-knowles:
- Introduce docker-compose file
- Correct certificate file names in spec config
- Push port expose to docker network to avoid conflict and ease setup,
also allow guest user to connect to rabbitmq "remotely" as docker won't
be on the loopback interface any more
- Allow peers to verify inside tests by setting up an alt name in the
certificate (currently all tests are set to not verify)

I did this work as part of getting setup for https://github.com/ruby-amqp/bunny/issues/440. It was more difficult than I had hoped to get setup so this hopefully helps with that.

It may also be of interest to use the docker compose setup in the [CI](https://docs.travis-ci.com/user/docker/#Using-Docker-Compose) then we could get rid of `unless ENV["CI"]` and have some of the tests using `verify_peer` on.


c-knowles:
I did this work as part of getting setup for https://github.com/ruby-amqp/bunny/issues/440. It was more difficult than I had hoped to get setup so this hopefully helps with that.

spec/tls/generate-server-cert.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+DIRNAME=$(dirname "$0")
+
+openssl req -sha256 -new -key $DIRNAME/server_key.pem -out $DIRNAME/server.csr -subj "/CN=mercurio" -config $DIRNAME/server-openssl.cnf
```
michaelklishin: You can potentially use [tls-gen](https://github.com/michaelklishin/tls-gen).

c-knowles:
It may also be of interest to use the docker setup in the CI then we could get rid of `unless ENV["CI"]` and have some of the tests using `verify_peer` on.

c-knowles:
I wanted to know if `spec/config/rabbitmq.config` is used from CI? I could not spot a usage but I left the file references as relative paths. If we can change that file freely then the `docker-compose.yml` also gets a bit neater.

michaelklishin:
No, Travis starts a RabbitMQ node managed via `sudo service`.

michaelklishin:
This doesn't have any side effects on the "local" workflow, so merging. We can make further corrections down the road.

PR #443, opened 2016-11-01
Support TLS SNI

c-knowles:
Add the hostname to the socket as per [the docs](ruby.github.io/openssl/OpenSSL/SSL/SSLSocket.html#method-i-hostname-3D).

Resolves https://github.com/ruby-amqp/bunny/issues/440.

Some of the tests depend on peer verification working. Locally this would be helped by https://github.com/ruby-amqp/bunny/pull/442 but the CI should still pass as the pre-existing blocks are ignored by CI. Perhaps afterwards we could get the CI to run all the tests using Travis' [Docker Compose support](https://docs.travis-ci.com/user/docker/#Using-Docker-Compose).


lib/bunny/transport.rb
```diff
@@ -364,6 +364,11 @@ def wrap_in_tls_socket(socket)
       raise "cannot wrap a socket into TLS socket, @tls_context is nil. This is a Bunny bug." unless @tls_context
 
       s = Bunny::SSLSocketImpl.new(socket, @tls_context)
+
+      # always set the SNI server name if possible since RFC 3546 and RFC 6066 both state
+      # that TLS clients supporting the extensions can talk to TLS servers that do not
+      s.hostname = @host if s.respond_to? :hostname
```
michaelklishin: Please use braces around method arguments.

spec/higher_level_api/integration/connection_spec.rb
```diff
@@ -336,13 +336,38 @@
...
+        let(:verify_peer) { false }
+
+        it "uses TLS port" do
+          expect(subject.port).to eq tls_port
+        end
+
+        it "sends the SNI details" do
+          # https://github.com/ruby-amqp/bunny/issues/440
+          subject.start
+          expect(subject.transport.socket.hostname).to eq("127.0.0.1")
```
michaelklishin: This fails when I run tests locally. Perhaps a better strategy would be checking if the hostname is in a set of `{"127.0.0.1", "::1", Socket.gethostname}`. And even that is likely to be fragile. So I'd probably remove this example entirely.

lib/bunny/transport.rb
c-knowles: I ran this through rubocop and it said to remove them, do you have a config I could try? (or any other similar tool)

c-knowles: Agreed, I just wanted to know what I could run it through to ensure anything I add is up to standard before I push it. Are you keen to have that checked in CI?

spec/higher_level_api/integration/connection_spec.rb
michaelklishin: Explicitly set it how? No, `RABBITMQ_URL` is not set.

c-knowles: I would probably set it by adding `:host` to the `described_class.new` parameters in this context only.

michaelklishin: It has the same fundamental proble : variability of host names. If you set it to an incorrect value for the environment then the transport will fail to connect ;)

lib/bunny/transport.rb
```diff
@@ -364,6 +364,11 @@ def wrap_in_tls_socket(socket)
       raise "cannot wrap a socket into TLS socket, @tls_context is nil. This is a Bunny bug." unless @tls_context
 
       s = Bunny::SSLSocketImpl.new(socket, @tls_context)
+
+      # always set the SNI server name if possible since RFC 3546 and RFC 6066 both state
+      # that TLS clients supporting the extensions can talk to TLS servers that do not
+      s.hostname = (@host) if s.respond_to? :hostname
```
michaelklishin: Sorry, what I had in mind is

``` ruby
s.hostname = @host if s.respond_to?(:hostname)
```

c-knowles: Ah, understood. Will get those few bits fixed up shortly.

michaelklishin: No opinion, using a setter works for me.

Issue #454, opened 2017-01-09
::Fixnum is deprecated in Ruby 2.4.0


carlhoerberg:
Since Ruby 2.4.0 Fixnum is deprecated. Starting Bunny in Ruby 2.4.0 gives the following warning: 

```
/app/vendor/bundle/ruby/2.4.0/gems/bunny-2.6.2/lib/bunny/channel.rb:1978: warning: constant ::Fixnum is deprecated
```

PR #458, opened 2017-01-20
Fixnum is deprecated

olleolleolle:
See #454.

(This is my first simple attempt to fix this issue.)


olleolleolle:
(It would of course be possible to rename AtomicFixnum, but to what end?)

carlhoerberg:
Won't cause any backward incompatiblies?

olleolleolle:
@carlhoerberg Integer is already the parent class of Fixnum.

PR #466, opened 2017-03-02
Fix #465

### LINKED ISSUE #465: create_channel on a connection intentionally closed by the user shown immediately throw an argument/illegal state exception
Hi, 

I just found this (weird?) behaviour: when creating a channel after a connection.close Bunny tries to recover from a network failure thus failing and getting into a Timeout exception. Is the client supposed to behave in this way? See attached code:

```ruby
require 'bunny'

client = Bunny.new
client.start

channel = client.create_channel
channel.default_exchange.publish "hello"

client.stop

client.create_channel

# Bunny tries to recover from a network failure

#<Bunny::Session:0x7fd0b8152300 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Recovering from a network failure...
#<Bunny::Session:0x7fd0b8152300 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Retrying connection on next host in line: 127.0.0.1:5672

# Timeout::Error (Timeout::Error)
```

Thanks.

madAle:
I provided some specs related to #465 and also suggested a fix that I hope correctly addresses the issue.

### DIFF
--- a/lib/bunny/exceptions.rb
+++ b/lib/bunny/exceptions.rb
@@ -97,9 +97,9 @@ module Bunny
     end
   end
 
-  class ConnectionManuallyClosed < Exception
+  class ConnectionAlreadyClosed < Exception
     def initialize
-      super('Connection has been manually closed. Call #start to reopen it')
+      super('Connection has been already closed')
     end
   end
 
--- a/lib/bunny/exceptions.rb
+++ b/lib/bunny/exceptions.rb
@@ -97,6 +97,12 @@ module Bunny
     end
   end
 
+  class ConnectionManuallyClosed < Exception
+    def initialize
+      super('Connection has been manually closed. Call #start to reopen it')
+    end
+  end
+
   class ShutdownSignal < Exception
   end
 
--- a/spec/issues/issue465_spec.rb
+++ b/spec/issues/issue465_spec.rb
@@ -25,26 +25,7 @@ describe Bunny::Session do
       it 'should raise an exception' do
         expect {
           connection.create_channel
-        }.to raise_error(Bunny::ConnectionManuallyClosed)
-      end
-    end
-
-    describe '#start' do
-      it 'should be possible to reopen it' do
-        connection.start
-        expect(connection.status).to eq :open
-      end
-    end
-
-    context 'and reopened' do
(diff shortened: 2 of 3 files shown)


spec/issues/issue465_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    describe '#create_channel' do
+      it 'should raise an exception' do
+        expect {
+          connection.create_channel
+        }.to raise_error(Bunny::ConnectionManuallyClosed)
+      end
+    end
+
+    describe '#start' do
+      it 'should be possible to reopen it' do
```
michaelklishin: I didn't plan on supporting this but if we do and it's actually a safe thing to do in practice, no objections from me.

lib/bunny/exceptions.rb
```diff
@@ -97,6 +97,12 @@ def initialize(frame)
     end
   end
 
+  class ConnectionManuallyClosed < Exception
```
michaelklishin: Other clients use `AlreadyClosedException` (Java) and similar names. I'd prefer that.

lib/bunny/exceptions.rb
```diff
@@ -97,6 +97,12 @@ def initialize(frame)
...
+    def initialize
+      super('Connection has been manually closed. Call #start to reopen it')
```
michaelklishin: I'm not sure how I feel about recommending this. It may or may not be safe and I don't think we should support reopening at all. It is hard and is not any easier to use than opening a new connection.
So let's drop the recommendation.

madAle: I agree.

spec/issues/issue465_spec.rb
madAle: Ok, given that I'll remove the recommendation given in current ConnectionManuallyClosed exception (name will change), I'll also drop these specs in order to maintain things coherent and not add official support for connection reopen. @michaelklishin do you agree?

michaelklishin: 👍

michaelklishin:
Thank you!

PR #493, opened 2017-03-30
Look up exchange by name in cache to match behavior described in method comment

jonbwhite:
The comments above the channel exchange declarations states:
```
Declares a fanout exchange or looks it up in the cache of previously declared exchanges.
```
However, the current code simply creates a new exchange every time, overwriting the value in the cache.

This pull request simply tries to look up the exchange by name in the cache and return it if found, and creates a new one otherwise, similarly to what is done in the `queue` method.


michaelklishin:
Thank you. This required correcting a test but otherwise looks identical to what we've been doing for queues for a very long time, so I consider this a positive change.

PR #495, opened 2017-04-03
Fix #492

### LINKED ISSUE #492: Windows cert file paths being treated like inline certs
I'm running v2.6.4 of the gem, and the code [here](https://github.com/ruby-amqp/bunny/blob/2.6.4/lib/bunny/transport.rb#L455) determines if each cert is a path or inline certificate by checking if the string starts with a `/`. This is fine for Linux, but on my Windows box with Ruby 2.3.1 (provided in the Chef DK), that list of certs is:
```ruby
[
  "C:/opscode/chefdk/embedded/ssl/certs/cacert.pem",
  "C:/opscode/chefdk/embedded/ssl/certs/ca-certificates.crt",
  "C:/opscode/chefdk/embedded/ssl/certs/ca-bundle.crt",
  "C:/opscode/chefdk/embedded/ssl/certs/ca-bundle.pem"
]
```

This causes these paths to be appended to the `cert_inlines` array, and eventually causes a `OpenSSL::X509::CertificateError (not enough data)` when it's added [here](https://github.com/ruby-amqp/bunny/blob/2.6.4/lib/bunny/transport.rb#L470).

I propose that the regex be updated from `/^\//` to `/^([a-z]:?)?\//i` to allow these Windows-style paths. I'd be glad to submit a PR to fix this, but wanted to run it by you guys first.
Thanks!

jsmartt:
Prevents Windows file paths from being treated as inline certs

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -51,6 +51,13 @@ Bunny now can work with frozen host arrays.
 GitHub issue: [#446](https://github.com/ruby-amqp/bunny/issues/446)
 
 
+### Support Windows SSL CA Cert File Paths
+
+Bunny now treats the default CA cert file paths as paths instead of inline certs.
+
+GitHub issue: [#492](https://github.com/ruby-amqp/bunny/issues/492)
+
+
 
 ## Changes between Bunny 2.5.0 and 2.6.0 (October 15th, 2016)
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you, please submit this against `2.6.x-stable`. Feel free to update the change log there but I can do it as well.

PR #499, opened 2017-04-21
Small fixes

madAle:
Hi, 

I've made these two small commits that address what I think are little oversights that I found in Session#recover_from_network_failure.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -718,7 +718,6 @@ module Bunny
         else
           @logger.debug "Not resetting recovery attempt counter after successful reconnection, as configured"
         end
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

michaelklishin:
@madAle I've cherry picked this to `2.6.x-stable`. Would it be acceptable for you to test these changes from that branch or do you need a tag and a preview gem release?

madAle:
@michaelklishin these are trivial "fixes" that I don't expect to have a big impact on code base. Anyhow I just manually tested them and it seems that everything is working as expected. 
In your opinion, is it worth to add some specs (or edit existent ones)?

PR #501, opened 2017-05-04
Remove block parameter from new and make run parameter options consistent with new.

meetme2meat:
- Removed unused block parameter 
    If I'm correct. I don't the see the `&block` parameter used. 

- Make `run` parameter options to be consistent with `new`


michaelklishin:
Thank you.

PR #503, opened 2017-05-11
Fixed the docs for the default socket timeouts

valo:
The default connection timeout is not 5 sec, but 30 sec: https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/transport.rb#L23

There are also 2 missing config options for read_timeout and write_timeout on the TCP socket. I added these: https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/transport.rb#L43


michaelklishin:
Thank you!

PR #511, opened 2017-07-18
0.7.x: support boolean headers

maxlaverse:
According to the specifications, it should be possible to provide boolean values for a header.
Currently the `Buffer.write` method does not support this types and lead to invalid packets.

This PR adds support for writing boolean header, and fixes the reading to return a boolean value.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,20 @@
-# https://github.com/travis-ci/travis-ci/wiki/.travis.yml-options
+language: ruby
 bundler_args: --without development
+cache: bundler
+before_script: "./bin/ci/before_build"
 script: "bundle exec rspec spec"
+dist: trusty
 rvm:
-  - 1.8.7
-  - ree
-  - 1.9.2
-  - 1.9.3
-  - rbx-2.0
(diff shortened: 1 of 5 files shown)


michaelklishin:
Thank you but I honestly don't know how I feel about doing a `0.7.x` release. What prevents you from upgrading?

skaes:
@michaelklishin we're running 50+ applications on this version, via our beetle gem. Upgrading would be quite some effort for us, and we have more important items on our roadmap.

I was planning to merge this pull request and release the new gem version, but you were quicker.

michaelklishin:
@skaes understood. Feel free to produce a release :) Will Beetle always use Bunny 0.7.x?

skaes:
@michaelklishin I don't think we'll always the old version. If we find the time, we might switch. But the new version is a big rewrite, so we have to be careful.

michaelklishin:
@skaes Bunny 0.9 was pretty compatible with 0.7.x but I understand what you mean. Bunny 0.7.x is the only AMQP 0-8 client I see used, though.

PR #523, opened 2017-10-24
Treat string and symbols the same when specifying 'server' or :server…

tyronewilson:
When loading config values from yaml file you can't easily pass in a symbol like :server. So  this PR treats 'server' the same as :server so that string values can be used from .yml files.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -1257,7 +1257,7 @@ module Bunny
 
     # @private
     def negotiate_value(client_value, server_value)
-      return server_value if client_value == :server
+      return server_value if [:server, "server"].include? client_value
 
       if client_value == 0 || server_value == 0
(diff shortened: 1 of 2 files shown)


lib/bunny/session.rb
```diff
@@ -1257,7 +1257,7 @@ def heartbeat_disabled?(val)
 
     # @private
     def negotiate_value(client_value, server_value)
-      return server_value if client_value == :server
+      return server_value if [:server, "server"].include? client_value
```
michaelklishin: Sorry to nitpick but can you please use brackets around the argument?

michaelklishin:
Also, this should go go into `2.7.x-stable`.

lib/bunny/session.rb
tyronewilson: sure I will do so.

PR #524, opened 2017-10-24
Treat string and symbols the same when specifying 'server' or :server

tyronewilson:
When loading config values from yaml file you can't easily pass in a symbol like :server. So  this PR treats 'server' the same as :server so that string values can be used from .yml files.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -1257,7 +1257,7 @@ module Bunny
 
     # @private
     def negotiate_value(client_value, server_value)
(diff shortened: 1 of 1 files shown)


tyronewilson:
I closed the other one as there were conflicts. This is the right one now.

tyronewilson:
Doh..... clicking too fast. Sorry

michaelklishin:
Thank you.

tyronewilson:
My pleasure :)

Issue #356, opened 2015-10-08
Cannot connect to RabbitMQ on JRuby 9.0.1.0 due to private method call Bunny::SocketImpl#open


naag:
I've just opened PR https://github.com/ruby-amqp/bunny/pull/355 to fix a issue where Bunny swallows a ```NoMethodError```. This is kind of a companion issue where I try to to report the root issue.

When running on JRuby 9.0.1.0 (other versions should be affected as well), Bunny cannot connect to RabbitMQ at all:

```
$ ruby -e 'require "bunny"; Bunny.new.start'
W, [2015-10-08T10:18:49.731000 #13626]  WARN -- #<Bunny::Session:0xfa4 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Could not establish TCP connection to 127.0.0.1:5672: 
Bunny::TCPConnectionFailedForAllHosts: Could not establish TCP connection to any of the configured hosts
  start at /home/buecker/.rvm/gems/jruby-9.0.1.0/gems/bunny-2.2.0/lib/bunny/session.rb:315
  <top> at -e:1
```

As you can see in my PR, the real issue behind this is a ```NoMethodError``` due to the way that ```Bunny::SocketImpl#open``` works on JRuby (https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/transport.rb#L279). This never happens on MRI because the modules are setup differently. I can't really explain it better, but the generous folks over at JRuby looked at this as well, maybe you want to checkout the conversation we had on IRC: http://logs.jruby.org/jruby/2015-10-07#14240751. Quoting:

```
21:50 <enebo> oh heh 
21:50 <enebo> cruby socket has self.open and jruby does not 
21:50 <headius> it will work in MRI because open is defined directly on Bunny::Socket 
21:50 <enebo> yeah 
21:51 <headius> enebo: right, but Bunny::JRuby::Socket includes Bunny::Socket 
21:51 <headius> but that doesn't pull in class methods 
```

I hope this makes sense to you and there's a way to improve this. If you need anything else, I'd be happy to help out!

michaelklishin:
Thank you but JRuby is no longer supported.
I'm happy to include your PR but don't count on JRuby support for the future. Simply use March Hare.

naag:
[...]

PR #529, opened 2017-11-16
Fix incompatibility with JRuby 9K - #356

dakull:
Currently bunny doesn't work at all with JRuby 9K - when trying to connect it fails with the behaviour described in #356.

This PR fixes that plus a couple of more issues, the integration specs are fully green but the full specs fail so this should be regarded as WIP + feedback on the changes.

Yes we have a valid option for JRuby (March Hare gem) but it would be nice for Bunny at least to function with JRuby 9K even if that implies a bigger overhead than native Java libs - Im thinking for some smaller apps having the full Bunny gem API might be useful.

### DIFF
--- a/lib/bunny/jruby/socket.rb
+++ b/lib/bunny/jruby/socket.rb
@@ -8,6 +8,20 @@ module Bunny
     module Socket
       include Bunny::Socket
 
+      def self.open(host, port, options = {})
+        socket = ::Socket.tcp(host, port, nil, nil,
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you. I will backport to `2.7.x`.

mach-kernel:
I have this same issue, thank you @dakull. Also, buna!

michaelklishin:
Released in `2.8.0`, will do a `2.7.3` later today.

PR #530, opened 2017-11-20
Do not log to error when automatically recover

nevesenin:
This prevents exceptions to be logged to error when automatically recover is enabled. I don't know if it should log in general in this case, but I think at least not to error.


michaelklishin:
Thank you for your time. I'm not a fan of not logging exceptions, including when automatic connection recovery is enabled.

nevesenin:
Ok, what about logging with a different log level, when when automatic connection recovery is enabled?

nevesenin:
#531

PR #531, opened 2017-11-20
Use log level warn instead of error ...

nevesenin:
... when automatic connection recovery is enabled

Our log management system creates alerts for messages with log level "error". When automatic connection recovery is enabled this doesn't feel right, because handling the exception is the expected behavior.


michaelklishin:
I have things to say about changing libraries due to the way a single company decided to set up log monitoring. However, using a warning when recovery is enabled is fine, so maybe I'll share it some other time.

PR #534, opened 2017-12-20
Handle URI Query Parameters 

Tensho:
Related Issues and PRs:
- ruby-amqp/amq-protocol#66
- ruby-amqp/amq-protocol#70
- ruby-amqp/amq-protocol#67
- ruby-amqp/amq-protocol#68
- ruby-amqp/amq-protocol#69.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -191,7 +191,7 @@ module Bunny
 
       client_props         = opts[:properties] || opts[:client_properties] || {}
       @client_properties   = DEFAULT_CLIENT_PROPERTIES.merge(client_props)
-      @mechanism           = normalize_auth_mechanism(opts.fetch(:auth_mechanism, "PLAIN"))
+      @mechanism           = (opts[:auth_mechanism] || []).first || "PLAIN"
       @credentials_encoder = credentials_encoder_for(@mechanism)
       @locale              = @opts.fetch(:locale, DEFAULT_LOCALE)
 
@@ -464,7 +464,7 @@ module Bunny
     # @param [String] uri amqp or amqps URI to parse
     # @return [Hash] Parsed URI as a hash
     def self.parse_uri(uri)
-      AMQ::Settings.parse_amqp_url(uri)
+      AMQ::Settings.configure(uri)
     end
 
     # Checks if a queue with given name exists.
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -303,12 +303,16 @@ module Bunny
       (opts[:port] == AMQ::Protocol::TLS_PORT) || false
     end
 
+    def tls_ca_certificates_paths_from(opts)
+      Array(opts[:cacertfile] || opts[:tls_ca_certificates] || opts[:ssl_ca_certificates])
+    end
+
     def tls_certificate_path_from(opts)
-      opts[:tls_cert] || opts[:ssl_cert] || opts[:tls_cert_path] || opts[:ssl_cert_path] || opts[:tls_certificate_path] || opts[:ssl_certificate_path]
+      opts[:certfile] || opts[:tls_cert] || opts[:ssl_cert] || opts[:tls_cert_path] || opts[:ssl_cert_path] || opts[:tls_certificate_path] || opts[:ssl_certificate_path]
     end
 
     def tls_key_path_from(opts)
-      opts[:tls_key] || opts[:ssl_key] || opts[:tls_key_path] || opts[:ssl_key_path]
+      opts[:keyfile] || opts[:tls_key] || opts[:ssl_key] || opts[:tls_key_path] || opts[:ssl_key_path]
     end
 
     def tls_certificate_from(opts)
@@ -337,7 +341,7 @@ module Bunny
     end
 
     def prepare_tls_context(opts)
-      if (opts[:verify_ssl] || opts[:verify_peer]).nil?
+      if (opts[:verify_ssl] || opts[:verify_peer] || opts[:verify]).nil?
         opts[:verify_peer] = true
       end
 
@@ -349,8 +353,8 @@ module Bunny
       @tls_key               = tls_key_from(opts)
       @tls_certificate_store = opts[:tls_certificate_store]
 
-      @tls_ca_certificates   = opts.fetch(:tls_ca_certificates, default_tls_certificates)
-      @verify_peer           = (opts[:verify_ssl] || opts[:verify_peer])
+      @tls_ca_certificates   = tls_ca_certificates_paths_from(opts) || default_tls_certificates
+      @verify_peer           = (opts[:verify_ssl] || opts[:verify_peer] || opts[:verify])
 
       @tls_context = initialize_tls_context(OpenSSL::SSL::SSLContext.new, opts)
     end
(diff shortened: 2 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -3,15 +3,10 @@
...
+Bundler.require(:test)
 
-
-require "effin_utf8"
 require "bunny"
-require "rabbitmq/http/client"
-
 
-require "amq/protocol/version"
 puts "Using Ruby #{RUBY_VERSION}, amq-protocol #{AMQ::Protocol::VERSION}"
```
Tensho: Maybe it's more relevant to show SHA of the `amq-protocol` master branch here instead of version.

spec/higher_level_api/integration/connection_spec.rb
```diff
@@ -319,9 +343,9 @@
           password: "bunny_password",
           vhost: "bunny_testbed",
           ssl:                 true,
-          ssl_cert:            "spec/tls/client_cert.pem",
```
Tensho: `:ssl` context test data setup is outdated, so I normalized it with `:tls` context,

spec/higher_level_api/integration/connection_spec.rb
```diff
@@ -46,6 +46,30 @@
...
       end
     end
+
+    context "when URI contains query parameters" do
+      subject(:session) do
+        described_class.new("amqps://bunny_gem:bunny_password@/bunny_testbed?heartbeat=10&connection_timeout=100&channel_max=1000&verify=false&cacertfile=spec/tls/ca_certificate.pem&certfile=spec/tls/client_certificate.pem&keyfile=spec/tls/client_key.pem")
+      end
+
+      it "parses extra connection parameters" do
+        # session.start # raises "OpenSSL::SSL::SSLError: hostname "127.0.0.1" does not match the server certificate"
```
Tensho: https://groups.google.com/d/msg/rabbitmq-users/V4fCdASYT1I/75TG_VhVCwAJ

lib/bunny/session.rb
```diff
@@ -464,7 +464,7 @@ def blocked?
     # @param [String] uri amqp or amqps URI to parse
     # @return [Hash] Parsed URI as a hash
     def self.parse_uri(uri)
-      AMQ::Settings.parse_amqp_url(uri)
+      AMQ::Settings.configure(uri)
```
Tensho: ❗️  Very important change that enables defaults merge for the parsed parameters.

Tensho:
Strange things happen at CI, I can't understand how `Bunny::Session#ssl` may return `0`?
https://travis-ci.org/ruby-amqp/bunny/builds/319109274
```
Failures:


  1) Bunny::Session initialized via connection URI when URI contains query parameters parses extra connection parameters

     Failure/Error: expect(session.ssl?).to eq(true)

     

       expected: true

            got: 0

     

       (compared using ==)

     # ./spec/higher_level_api/integration/connection_spec.rb:66:in `block (4 levels) in <top (required)>'


Finished in 3 minutes 0 seconds (files took 0.2697 seconds to load)

198 examples, 1 failure


Failed examples:


rspec ./spec/higher_level_api/integration/connection_spec.rb:55 # Bunny::Session initialized via connection URI when URI contains query parameters parses extra connection parameters
```

spec/spec_helper.rb
michaelklishin: What is there is no "SHA" because we don't run with a local clone?

spec/higher_level_api/integration/connection_spec.rb
michaelklishin: I responded there.

michaelklishin:
I get several TLS connection test suite failures in my standard testing environment, will take a look later in the week.

lib/bunny/session.rb
```diff
@@ -191,7 +191,7 @@ def initialize(connection_string_or_opts = ENV['RABBITMQ_URL'], optz = Hash.new)
 
       client_props         = opts[:properties] || opts[:client_properties] || {}
       @client_properties   = DEFAULT_CLIENT_PROPERTIES.merge(client_props)
-      @mechanism           = normalize_auth_mechanism(opts.fetch(:auth_mechanism, "PLAIN"))
+      @mechanism           = (opts[:auth_mechanism] || []).first || "PLAIN"
```
michaelklishin: If we modify this at all (not sure why it's a part of this PR but OK), `normalize_auth_mechanism` should then be deleted.

spec/higher_level_api/integration/connection_spec.rb
michaelklishin: Inspecting the transport suggests that it has `:ssl=>0`.

michaelklishin:
So far everything points at the fact that `@opts` in `Bunny::Session` in your test have `:ssl => 0`. I pushed `Tensho-init-params` with my changes that were necessary to make the tests except for one (the same as yours) pass for me.

Please open a new PR based on that branch once you figure out what's going on in the newly added test example. Thanks.

michaelklishin:
So the issue comes from ``:

``` ruby
require "pp"

pp AMQ::Settings.configure("amqps://bunny_gem:bunny_password@/bunny_testbed?heartbeat=10&connection_timeout=100&channel_max=1000&verify=false&cacertfile=spec/tls/ca_certificate.pem&certfile=spec/tls/client_certificate.pem&keyfile=spec/tls/client_key.pem")
```

produces

```
{:host=>"127.0.0.1",
 :port=>5671,
 :user=>"bunny_gem",
 :pass=>"bunny_password",
 :vhost=>"bunny_testbed",
 :ssl=>0,
 :frame_max=>131072,
 :heartbeat=>10,
 :connection_timeout=>100,
 :channel_max=>1000,
 :auth_mechanism=>nil,
 :verify=>"false",
 :fail_if_no_peer_cert=>nil,
 :cacertfile=>"spec/tls/ca_certificate.pem",
 :certfile=>"spec/tls/client_certificate.pem",
 :keyfile=>"spec/tls/client_key.pem",
 :scheme=>"amqps"}
```

lib/bunny/session.rb
Tensho: Yeah, nice spot, I forgot to remove the method. Basically `auth_mechanism` confused me too. According to specification client may specify several auth mechanism, so I decided to adheer in `amq-protocol`. But when I proceed with `bunny`, I realised that session initialization expects a single string value, not array. TBH, I didn't have time to analyze, how several auth mechanisms should be handled properly, so i decided to adheer bunny fallback to `"PLAIN"` for array case.

Let me recheck, maybe `normalize_auth_mechanism` is OK for array default value, I don't remember exact reason, why I replaced it...

spec/spec_helper.rb
michaelklishin: We can do that in `.travis.yml`.

lib/bunny/transport.rb
```diff
@@ -349,8 +353,8 @@ def prepare_tls_context(opts)
       @tls_key               = tls_key_from(opts)
       @tls_certificate_store = opts[:tls_certificate_store]
 
-      @tls_ca_certificates   = opts.fetch(:tls_ca_certificates, default_tls_certificates)
-      @verify_peer           = (opts[:verify_ssl] || opts[:verify_peer])
+      @tls_ca_certificates   = tls_ca_certificates_paths_from(opts) || default_tls_certificates
```
carlhoerberg: I susptect that `tls_ca_certificates_paths_from` will never evaluate to nil, so the `default_tls_certificates` will never be used

Tensho: You are totally right, what do you think if we call `Array()` out of `#tls_ca_certificates_paths_from` method?
```
@tls_ca_certificates   = Array(tls_ca_certificates_paths_from(opts) || default_tls_certificates)
```
Should I create a PR with fix?

carlhoerberg: yes, great

PR #535, opened 2017-12-21
Extend URI Parsing Specs 

Tensho:
Follow up to #534.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     container_name: bunny_rabbitmq
     environment:
       RABBITMQ_NODENAME: bunny
+      BUNNY_RABBITMQ_HOSTNAME: mercurio
       # link to spec specific configuration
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
     container_name: bunny_rabbitmq
     environment:
       RABBITMQ_NODENAME: bunny
+      BUNNY_RABBITMQ_HOSTNAME: mercurio
```
Tensho: Not sure what should be chosen as default. Please check my considerations in Google Groupe.
https://groups.google.com/d/msg/rabbitmq-users/V4fCdASYT1I/J5ss8yCzCwAJ

michaelklishin:
Please never ever hardcode directories and values. Yes, there is a directory with generated certificates used as example. Even I cannot use it from every machine because not all of them resolve `mercurio`. There are two methods that return current hostname and tls-gen profile result/certificate/key directory. They are already used in some tests.

Issue #541, opened 2018-01-23
Problem with verify_peer: false


hading:
Going from Bunny 2.7.0 to 2.9.1 I found that verify_peer: false was no longer being respected when in my configuration options for a connection. I think I have a handle on the problem.

From an irb session with 2.9.1:

```
irb(main):028:0> b = Bunny.new(ssl: true, host: 'rabbitmq.host', user: 'user', password: 'password', vhost: 'downloader', verify_peer: false)
W, [2018-01-23T17:32:08.049997 #34207]  WARN -- #<Bunny::Session:0x7f8d4a1fe698 @rabbitmq.host:5671, vhost=, addresses=[rabbitmq.host:5671]>: Using TLS but no client certificate is provided! If RabbitMQ is configured to verify peer
certificate, connection upgrade will fail!

=> #<Bunny::Session:0x7f8d4a1fe698 user@rabbitmq.host:5671, vhost=downloader, addresses=[rabbitmq.host:5671]>
irb(main):029:0> t = b.transport
[...]

PR #542, opened 2018-01-24
Set default value for opts[:verify_peer] to true only when all of the keys checked are nil.

hading:
This is for issue #541 which I filed.

It appears that this bit of code is intended to force peer verification when none of the appropriate options have been provided. 

However, the existing conditional does not do this. For example, if in the options we pass in verify_peer: false and leave the other two blank then we evaluate:

`
(nil || false || nil).nil?
`

which is actually true and winds up setting `opts[:verify_peer] = true`, which is unlikely to be what we wanted when we explicitly passed in a value of false. (In fact here the first two can independently be any combination of nil and false and the third one will determine whether the parenthesized piece is nil or false.)

This modification only sets the default when all three of these keys are actually nil.

### DIFF
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -341,7 +341,7 @@ module Bunny
     end
 
     def prepare_tls_context(opts)
-      if (opts[:verify_ssl] || opts[:verify_peer] || opts[:verify]).nil?
+      if opts.values_at(:verify_ssl, :verify_peer, :verify).all?(&:nil?)
         opts[:verify_peer] = true
       end
 


michaelklishin:
The CI failure is a Bundler/Ruby HEAD compatibility problem.

michaelklishin:
@Tensho @carlhoerberg FYI.

michaelklishin:
@hading thank you! I back ported this to `2.9.x-stable`. Any chance you can try it for a few days before we cut a `2.9.2`?

hading:
Sure, no problem.

Issue #519, opened 2017-09-23
Can't handle heartbeat 0 from server


carlhoerberg:
rabbitmq.config:
```
[
  {rabbit, [
    {heartbeat, 0}
  ]}
].
```

client:
```
require 'bunny'
Bunny.new("amqp://guest:guest@localhost").start
```

exception:
```
E, [2017-09-23T11:10:23.309983 #41848] ERROR -- #<Bunny::Session:0x7f7fd9029490 guest@localhost:5672, vhost=/, addresses=[localhost:5672]>: Got an exception when receiving data: IO timeout when reading 7 bytes (Timeout::Error)
W, [2017-09-23T11:10:23.310777 #41848]  WARN -- #<Bunny::Session:0x7f7fd9029490 guest@localhost:5672, vhost=/, addresses=[localhost:5672]>: An empty frame was received while opening the connection. In RabbitMQ <= 3.1 this could mean an authentication issue.
/Users/carl/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bunny-2.7.0/lib/bunny/session.rb:327:in `rescue in start': Could not establish TCP connection to any of the configured hosts (Bunny::TCPConnectionFailedForAllHosts)
        from /Users/carl/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bunny-2.7.0/lib/bunny/session.rb:295:in `start'
        from test/ruby.rb:2:in `<main>'
```

michaelklishin:
I am inclined to treat this as a feature: disabling heartbeats is a terrible idea.

carlhoerberg:
Not if TCP keep alive is enabled, then it simplifies things a lot.

michaelklishin:
A quick look doesn't make it obvious what's going on: there's a default read timeout on the socket and heartbeat sender isn't started after connection is negotiated.

However, Wireshark reveals it's the client that closes the connection:

```
49	16.680471	127.0.0.1	127.0.0.1	TCP	56	65110 → 5672 [FIN, ACK] Seq=399 Ack=529 Win=407776 Len=0 TSval=953851427 TSecr=953851427
```

carlhoerberg:
May it be that Bunny sets its TCP read timeout to 0 when the heartbeat is
0?

michaelklishin:
@carlhoerberg that was my first guess. So far I'm unable to prove that hypothesis.

michaelklishin:
[...]

PR #551, opened 2018-04-26
If heartbeats are disabled, do not use a read timeout when reading frame header

carlhoerberg:
Bug #519 wasn't quite resolved. If heartbeats are disabled (keepalive
used instead), Bunny would use the default read_timeout when waiting for
next frame, that meant that the connection would be closed with an
exception if there was no activity for 30s.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -1177,22 +1177,11 @@ module Bunny
       @logger.debug { "Heartbeat interval negotiation: client = #{@client_heartbeat}, server = #{connection_tune.heartbeat}, result = #{@heartbeat}" }
       @logger.info "Heartbeat interval used (in seconds): #{@heartbeat}"
 
-      # We set the read_write_timeout to twice the heartbeat value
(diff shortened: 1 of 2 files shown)


carlhoerberg:
Yes, this disables heartbeat timeouts as well as they're intertwined. Suggestions welcomed on how to easiest fix that.

carlhoerberg:
Disabled read_timeout completely now when heartbeat is disabled too. The behavior of connections with heartbeat enabled is unchanged with this commit.

michaelklishin:
Thanks, I will QA this fairly soon. Probably worth bumping a minor version for if we end up merging it.

michaelklishin:
After giving it some thought I think it makes sense to disable TCP socket read timeouts if heartbeats are disabled. The two are interconnected in practice. I left a note about TCP keepalives and a link to [RabbitMQ heartbeats guide](http://www.rabbitmq.com/heartbeats.html) in the change log.

Thank you, @carlhoerberg.

PR #555, opened 2018-06-05
 Use OpenSSL::X509::Store#set_default_paths

Ana06:
Use `OpenSSL::X509::Store#set_default_paths` to add default certificates instead of reimplementing this method in `Transport#default_tls_certificates`. :bowtie: See: https://ruby-doc.org/stdlib-2.5.0/libdoc/openssl/rdoc/OpenSSL/X509/Store.html#method-i-set_default_paths

`Transport#default_tls_certificates`, appart from the work to maintain the code, had several problems:
- It didn't work for JRuby
- It didn't work for openSUSE, as `ca-bundle.pem` doesn't exist anymore

### DIFF
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -357,7 +357,6 @@ module Bunny
       @tls_key               = tls_key_from(opts)
       @tls_certificate_store = opts[:tls_certificate_store]
 
-      @tls_ca_certificates   = tls_ca_certificates_paths_from(opts)
       @verify_peer           = as_boolean(opts[:verify_ssl] || opts[:verify_peer] || opts[:verify])
 
       @tls_context = initialize_tls_context(OpenSSL::SSL::SSLContext.new, opts)
@@ -416,7 +415,7 @@ module Bunny
       ctx.cert_store = if @tls_certificate_store
                          @tls_certificate_store
                        else
-                         initialize_tls_certificate_store(@tls_ca_certificates)
+                         initialize_tls_certificate_store(tls_ca_certificates_paths_from(opts))
                        end
 
       if !@tls_certificate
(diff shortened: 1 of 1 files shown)


michaelklishin:
That's a nice find. The method seems to be available even in Ruby 2.0. I'm testing it as we speak on a few platforms. Thank you!

michaelklishin:
It turns out the paths are not ignored, as in, they are added to the store as far as I can tell (Ruby's OpenSSL API doesn't make it easy to verify). What is broken now is the `tls_ca_certificates` reader since there no longer is an ivar.

michaelklishin:
Thank you, it's definitely nice to have OpenSSL set up its own paths as opposed to trying to track them down.

This is a potentially breaking change that's hard to validate due to the number of platform/Ruby version permutations that would need testing.

So the earliest we can release it in is `2.11.0`. Which can still happen relatively soon.

Ana06:
:tada: 

> So the earliest we can release it in is 2.11.0. Which can still happen relatively soon.

A release with the change would be nice :wink:

Ana06:
@michaelklishin when do you plan to release 2.11.0? Is there something else to do so before that happens? Maybe I can help so that it is released faster :wink:

michaelklishin:
`2.10.0` shipped yesterday and hasn't been announced yet. I can produce a preview release any time.

Ana06:
@michaelklishin 

> I can produce a preview release of 2.11.0 any time.

that would be really appreciated :pray:

michaelklishin:
@Ana06 [done](https://rubygems.org/gems/bunny/versions/2.11.0.pre1).

Ana06:
@michaelklishin thanks! :cupid:

michaelklishin:
`2.11.0` is up on RubyGems.

Issue #549, opened 2018-04-11
No failure when exceeding `recovery_attempts`


robinroestenburg:
**Reproduction scenario:**
* Setup a Bunny session using the following configuration parameters:
  * `automatically_recover: true`
  * `recovery_attempts: 5`
  * `network_recovery_internal: 5`
* Send a couple of messages to RabbitMQ using this session.
* Stop RabbitMQ server.
* Bunny starts the network recovery and retries the connection for 5 times. After that it gives up with the message `"Ran out of recovery attempts"`.  
No exception is raised by Bunny.
* Send more messages to RabbitMQ.  
No exception is generated by Bunny.

**Expectation:**
Error/exception is generated when either:
* network recovery fails after the maximum number of attempts, or
* messages are sent using a Bunny session that ran out of retry attempts.

---

**Cause**
See the following lines: https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L745-L756

If a network failure happens the exception is caught and the retry mechanism kicks in. When the maximum number of recovery attempts has been reached, the `should_retry_recovery?` method will return `false` and Bunny will swallow the exception.

This is a problem, because:
* the client code is not informed (through an exception or maybe a special status on the session object) that a problem occurred, and
* any future messages will also be silently dropped.

The original commit message for the `recovery_attempts` feature can be found here: https://github.com/ruby-amqp/bunny/commit/fff7f4465d2261e8bfbb840d9012c8a13edb9929
It states:

> `:recovery_attempts` options sets the maximum number of attempts Bunny
would make to recover from network failures **before giving up and
throwing the original exception**. `:recovery_attempts` is effective only
if `:automatically_recover` option is true.

It is not re-raising the original exception right now. I tried to add it locally, but that didn't seem to work. Maybe you have some ideas?

If you need any more information please let me know.

michaelklishin:
Feel free to submit a PR.

Connections that are out of retry attempts should transition to the closed state, which will automatically lead to any operation on it throw an exception.

robinroestenburg:
@michaelklishin ok, sounds good. I will work on a PR.

michaelklishin:
Note that connection recovery happens in the I/O loop thread. But setting an ivar that would stop recovery attempts from happening should be enough and only the I/O loop thread should mutate it (at least in theory).

robinroestenburg:
[...]

PR #556, opened 2018-07-27
Issue #549: Move connection to closed state when recovery attempts fail

arlandism:
See #549 for a more thorough description of the problem.

A gist:

When the RabbitMQ server goes down, the session attempts to continually
reconnect, only stopping once the recovery_attempts threshold is met (or
forever is this parameter is not provided). In the former case, there
are no exceptions raised to indicate to clients that new messages are
not being processed. This patch provides that by initiating close
from the session, which will close related channels and the reader
loop. Then, an exception will be thrown when subsequent operations are
performed.

Client code will now be able to rescue `Bunny::ConnectionClosedError` to gracefully degrade.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -757,6 +757,7 @@ module Bunny
         end
       else
         @logger.error "Ran out of recovery attempts (limit set to #{@max_recovery_attempts})"
+        self.close
       end
     end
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
This solution makes a lot of sense. Thank you!

PR #557, opened 2018-08-02
Fix comments on exchange_declare

BrunoMVPCosta:
The comments had echange instead of exchange and the link to the
"exchanges and publishing guide" was wrong as well.


michaelklishin:
Thank you!

Issue #560, opened 2018-08-13
Closed transports cause silent failures on writes


sbonebrake:
Reproduced with Bunny 2.11.0 and RabbitMQ 3.7.6

If there is a closed transport, transport#write will silently fail when attempting to publish.  This is when using automatically_recover: true and threaded: false.

Code to reproduce this issue can be found here: https://github.com/sbonebrake/bunny_repro/blob/master/heartbeat_failure_causes_silent_write_fails.rb

```
D, [2018-08-13T12:54:40.696182 #2936] DEBUG -- : Sent protocol preamble
D, [2018-08-13T12:54:40.701428 #2936] DEBUG -- : Sent connection.start-ok
D, [2018-08-13T12:54:40.702059 #2936] DEBUG -- : Heartbeat interval negotiation: client = 3, server = 60, result = 3
I, [2018-08-13T12:54:40.702093 #2936]  INFO -- : Heartbeat interval used (in seconds): 3
D, [2018-08-13T12:54:40.702128 #2936] DEBUG -- : Will use socket read timeout of 6.6000000000000005
D, [2018-08-13T12:54:40.702145 #2936] DEBUG -- : Initializing channel ID allocator with channel_max = 2047
D, [2018-08-13T12:54:40.702249 #2936] DEBUG -- : Sent connection.tune-ok with heartbeat interval = 3, frame_max = 131072, channel_max = 2047
D, [2018-08-13T12:54:40.702347 #2936] DEBUG -- : Sent connection.open with vhost = /
D, [2018-08-13T12:54:40.703017 #2936] DEBUG -- : Initializing heartbeat sender...
D, [2018-08-13T12:54:40.703150 #2936] DEBUG -- : Allocated channel id: 1
D, [2018-08-13T12:54:40.704355 #2936] DEBUG -- : Session#handle_frame on 1: #<AMQ::Protocol::Channel::OpenOk:0x00007f83ff2fb798 @channel_id="">
I, [2018-08-13T12:54:40.704409 #2936]  INFO -- : Declaring Bob with ch.queue('Bob')...
D, [2018-08-13T12:54:40.705170 #2936] DEBUG -- : Session#handle_frame on 1: #<AMQ::Protocol::Queue::DeclareOk:0x00007f83ff2fa848 @queue="Bob", @message_count=0, @consumer_count=0>
[...]

PR #562, opened 2018-08-14
send_frameset can now raise ConnectionClosedErrors

### LINKED ISSUE #561: Ensure that a connection that detects connectivity issues is transitioned into the "closed" state
Currently when a heartbeat is detected, `Bunny::Session` could be left in a state other than "closed".
Even if connection recovery is underway, the correct state for such connection is nonetheless "closed".

With this change `Bunny::Session#send_frameset_without_timeout` and friends could then throw a `ConnectionClosedError` which would be visible to clients. This is closer to how some other clients (e.g. Java) work.

@acogoluegnes @lukebakken FYI.

sbonebrake:
Fixes #561

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,7 @@ end
 group :test do
   gem "rspec", "~> 3.5.0"
   gem "rabbitmq_http_api_client", "~> 1.9.1", require: "rabbitmq/http/client"
+  gem "toxiproxy", "~> 1.0.3"
 end
 
 gemspec
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,3 +18,11 @@ services:
       - 15672:15672
     volumes:
       - ./spec:/spec:ro
+  toxiproxy:
+    container_name: toxiproxy
+    image: shopify/toxiproxy
+    ports:
+      - 8474:8474
+      - 11111:11111
+    depends_on:
+      - rabbitmq
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -1079,9 +1079,13 @@ module Bunny
       # still recommend not sharing channels between threads except for consumer-only cases in the docs. MK.
       channel.synchronize do
         # see rabbitmq/rabbitmq-server#156
-        data = frames.reduce("") { |acc, frame| acc << frame.encode }
-        @transport.write(data)
-        signal_activity!
+        if open?
+          data = frames.reduce("") { |acc, frame| acc << frame.encode }
+          @transport.write(data)
+          signal_activity!
+        else
+          raise ConnectionClosedError.new(frames)
+        end
       end
     end # send_frameset(frames)
 
@@ -1097,8 +1101,12 @@ module Bunny
       # If we synchronize on the channel, however, this is both thread safe and pretty fine-grained
       # locking. See a note about "single frame" methods in a comment in `send_frameset`. MK.
       channel.synchronize do
-        frames.each { |frame| self.send_frame_without_timeout(frame, false) }
-        signal_activity!
+        if open?
+          frames.each { |frame| self.send_frame_without_timeout(frame, false) }
+          signal_activity!
+        else
+          raise ConnectionClosedError.new(frames)
+        end
       end
     end # send_frameset_without_timeout(frames)
 
--- a/spec/higher_level_api/toxiproxy/basic_publish_toxi_spec.rb
+++ b/spec/higher_level_api/toxiproxy/basic_publish_toxi_spec.rb
@@ -1,0 +1,31 @@
+require "spec_helper"
+require_relative "./toxiproxy_helper"
+
+describe Bunny::Channel, "#basic_publish" do
+  before(:all) do
+    setup_toxiproxy
+    @connection = Bunny.new(:user => "bunny_gem", :password => "bunny_password", :vhost => "bunny_testbed",
+      host: "localhost:11111", heartbeat_timeout: 1)
+    @connection.start
+  end
+
+  after :all do
+    @connection.close if @connection.open?
+  end
+
+  context "when the the connection isn't open" do
+    let(:queue_name) { "bunny.basic.publish.queue#{rand}" }
+
+    it "raises a ConnectionClosedError" do
+      ch = @connection.create_channel
+      begin
+        rabbitmq_toxiproxy.down do
+          sleep 2
+          expect { ch.default_exchange.publish("", :routing_key => queue_name) }.to raise_error(Bunny::ConnectionClosedError)
+        end
+      ensure
+        cleanup_toxiproxy
+      end
+    end
+  end
+end
--- a/spec/higher_level_api/toxiproxy/toxiproxy_helper.rb
+++ b/spec/higher_level_api/toxiproxy/toxiproxy_helper.rb
@@ -1,0 +1,16 @@
+def setup_toxiproxy
+  Toxiproxy.populate([{
+    name: "rabbitmq",
+    listen: "0.0.0.0:11111",
+    upstream: "rabbitmq:5672"
+  }])
+  rabbitmq_toxiproxy.enable
+end
+
+def cleanup_toxiproxy
+  Toxiproxy.populate()
+end
+
+def rabbitmq_toxiproxy
+  Toxiproxy[/rabbitmq/]
+end


sbonebrake:
@michaelklishin Is this the approach you had in mind? If so I'll add some test cases.

michaelklishin:
Thank you. That's one half of what I had in mind. We also need to make missed heartbeats and I/O errors to first mark connection as closed (or maybe introduce a new state, `recovering`, that would still be different from `open`?).

Test cases would be great, we can adopt Toxiproxy for these new tests.

sbonebrake:
I think it  a new state for ```recovering``` is a good idea because right now ```@recovering_from_network_failure``` isn't protected by a mutex.

```ruby
    def handle_network_failure(exception)
      raise NetworkErrorWrapper.new(exception) unless @threaded

      @status_mutex.synchronize { @status = :disconnected }

      if !recovering_from_network_failure?
        begin
          @recovering_from_network_failure = true
```

Is it intended to have any thread be able to start the recovery? It would likely be the the reader_loop thread, but it seems like it could be heartbeat or the session thread. This is potentially bad if it happens when I make any call to rabbitmq. Eg ```session.create_channel()``` could hit a write_timeout, but then end up in the ```recover_from_network_failure``` method and sleep up to ```@network_recovery_interval * @max_recovery_attempts``` But maybe that's okay because it was initialized with automatically_recover?

michaelklishin:
Recovery happens in the I/O loop thread and there's only one per connection.

michaelklishin:
Extra synchronisation can be added but in 6 or so years of the current architecture of Bunny concurrent recovery attempts hasn't been an issue.

sbonebrake:
Couldn't this be a possible call stack?
```
Session#create_channel
Channel#open
Session#open_channel
Transport#send_frame
Transport#write    #Hits Timeout::Error
Session#handle_network_failure
Session#recover_from_network_failure
```
Maybe I'm missing something. I can try to get a repro of that with toxiproxy another time.  
As for ```reconnecting``` state, same PR or different one to keep them small?

michaelklishin:
Connections are generally not shared in Ruby apps (that hardly have any concurrency that developers intentionally introduce). I don't mind related changes in a single PR.

spec/lower_level_api/integration/basic_publish_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    @connection.close if @connection.open?
+  end
+
+  context "when the the connection isn't open" do
+    let(:queue_name) { "bunny.basic.publish.queue#{rand}" }
+
+    it "raises a ConnectionClosedError" do
+      ch = @connection.create_channel
+      # Assume the IO thread closed the connection due to exception
+      @connection.instance_variable_set("@status", :closed)
```
sbonebrake: Is the test case fine like this? or would you prefer a higher level toxiproxy test?

michaelklishin: You can simply close the connection to the same effect. A Toxiproxy test would be more useful.

sbonebrake: If you close the connection with session#close you would get the other exception ```Bunny::ChannelAlreadyClosed:
       cannot use a closed channel! Channel id: 1``` 
because channel knows it's closed. I'll look into a toxiproxy test case. Remove this one?

sbonebrake:
You mentioned "We also need to make missed heartbeats and I/O errors to first mark connection as closed". Both heartbeat thread and reader_loop will call handle_network_failure from the Transport class. With automatic_recovery the transport will be marked as closed and the session status will be ```:disconnected```. Is that what you mean? Or do you want ```:disconnected``` changed to ```:closed``? Without automatic_recovery the transport will be closed and the session thread will get an exception raised on it, but the status will still be ```:connected```.

michaelklishin:
@sbonebrake your understanding of the issue is mostly correct. I do not want to focus on the status of the transport since the transport is both an implementation detail and a disposable (as in, you have to instantiate a new one) dependency. `Bunny::Session` state is what I had in mind.

There's a difference between `:disconnected` and `:closed` that's also present in other clients (but not as a state). The former means "this connection is no longer open" while the latter tries to indicate "this connection was explicitly closed by the user". In Java client there is a "shutdown reason" to help distinguish between the two (in fact, three in that client) scenarios.

As far as `Bunny::Session#open?` goes they both should be identical. `:disconnected` therefore makes more sense for missed heartbeats.

Of course you can argue a case for a different set of states or just `:closed`. As long as we are making progress towards the idea in #560 which

michaelklishin:
So the behavior we are looking for (which partially might already be the case):

 * When a connection is explicitly closed with `Bunny::Session#close`, publish attempts on it throw a meaningful exception
 * When a connection is closed remotely but cleanly (e.g. via the HTTP API, see existing connection recovery tests), publish attempts on it throw a meaningful exception
 *  * When a connection is closed due to missed heartbeats, publish attempts… well, you get the idea ;)

michaelklishin:
I found something in commit logs (0e773475adb5bc6601dfb4fa4b709655ddf44570) that makes me think that Bunny has been pretty consistent about `:disconnected` vs `:closed`. When a connection is closed, it will never be recovered. When it is disconnected, that condition is temporary. 0e773475adb5bc6601dfb4fa4b709655ddf44570 transitions connections to `:closed` when they run out of recovery attempts.

sbonebrake:
Good to know, makes sense to me. Is ```:disconnected``` any different from ```:not_connected```? Session#close is interesting because it first sets status to ```:not_connected``` via close_connection then sets it to ```:closed```

michaelklishin:
@sbonebrake no, just an unintentional inconsistency. Connection closure is a request-response kind of operation. Other clients use `:closing` or similar. It makes a bit more sense to me than `:disconnected` in this particular case.

sbonebrake:
Ok, you mentioned three scenarios here:


>  When a connection is explicitly closed with Bunny::Session#close, publish attempts on it throw a meaningful exception

This is covered by channel_close_spec.rb
>  When a connection is closed remotely but cleanly (e.g. via the HTTP API, see existing connection recovery tests), publish attempts on it throw a meaningful exception

Covered by connection recovery tests

> . 3. When a connection is closed due to missed heartbeats, publish attempts…
Covered by new basic_publish_toxi_spec.rb

Is there anything else I should cover in this PR?  Happy to fix any style/naming preferences you have as well.

michaelklishin:
@sbonebrake please move the Toxiproxy tests under `spec/higher_level/integration`. The helper file can be folded into `spec_helper.rb` in a module.

I'd like to have a way to have them skipped when Toxiproxy isn't available (e.g. when you are not running the dependencies in Docker and don't have it running locally). We also make some tests that require setup steps conditional (e.g. TLS). Toxiproxy is a good candidate for that. As far as I can tell in the Ruby client `Toxiproxy.running?` is the predicate we want.

michaelklishin:
I tried running Toxiproxy locally via Homebrew with all defaults and it logs the following:

```
INFO[0000] API HTTP server starting                      host=localhost port=8474 version=2.1.2
INFO[0009] Started proxy                                 name=rabbitmq proxy=[::]:11111 upstream=rabbitmq:5672
INFO[0009] Accepted client                               client=[::1]:62399 name=rabbitmq proxy=[::]:11111 upstream=rabbitmq:5672
ERRO[0009] Unable to open connection to upstream         client=[::1]:62399 name=rabbitmq proxy=[::]:11111 upstream=rabbitmq:5672
```

even though a RabbitMQ node is running doesn't log anything. This likely means that RabbitMQ closes TCP connection that sends no data after a handshake period but also doesn't log it since, well, it sent no data (this is to avoid log spam from TCP load balancer health checks). I haven't done a traffic capture to further confirm this. Are there any manual setup steps that I'm missing?

michaelklishin:
@sbonebrake never mind, I'm happy to make those largely cosmetic changes. All tests but TLS pass with Docker Compose but that is due to either stale certificates or recent tls-gen changes that break on MacOS with LibreSSL.

michaelklishin:
@sbonebrake thank you! Let me know how soon you need a preview release up on rubygems.org.

sbonebrake:
Thanks @michaelklishin Let me if you're still having trouble with toxiproxy. I didn't really try to run it without docker-compose.
A preview release soon would be nice, but for now I can always pin my Gemfile to a git sha in the interim.

Issue #558, opened 2018-08-03
channel#queue can return wrong queue after Timeout::Error


sbonebrake:
Reproduced with Bunny 2.11.0 and RabbitMQ 3.7.6

**channel#queue will return wrong queue if called after a Timeout::Error occurs during previous #queue call.**

We hit this issue where we started getting messages enqueued in the wrong queues. In our logs we found a Timeout::Error with this call stack:
```
gems/bunny-2.11.0/lib/bunny/concurrent/continuation_queue.rb:39:in `block in poll'
gems/bunny-2.11.0/lib/bunny/concurrent/continuation_queue.rb:30:in `synchronize'
gems/bunny-2.11.0/lib/bunny/concurrent/continuation_queue.rb:30:in `poll'
gems/bunny-2.11.0/lib/bunny/channel.rb:1780:in `wait_on_continuations'
gems/bunny-2.11.0/lib/bunny/channel.rb:1023:in `queue_declare'
gems/bunny-2.11.0/lib/bunny/queue.rb:372:in `declare!'
gems/bunny-2.11.0/lib/bunny/queue.rb:51:in `initialize'
gems/bunny-2.11.0/lib/bunny/channel.rb:427:in `new'
gems/bunny-2.11.0/lib/bunny/channel.rb:427:in `queue'
```

Our code does not dispose of the Session when this occurs. In this case, the next call to Bunny is another channel#queue for a different queue. During this second call, the wait_on_continuations call returns the DeclareOk from the first call. This causes the first queue to be incorrectly returned from channel#queue, but the channels \@queues will contain the correct mapping.

**Order of events:**
```
1. Initialize with Bunny.new(url) with all defaults
2. channel = connection.create_channel
3. channel.queue("Bob")
4. Wait 15 Seconds (DEFAULT_CONTINUATION_TIMEOUT) for Timeout::Error from ContinuationQueue#poll
5. App catches Timeout::Error and logs it
6. AMQP returns to reader loop with DeclareOk-Bob sometime after 15s. This pushes on to ContinuationQueue's @q
7. channel.queue("Sally")
8. channel calls wait_on_continuations 
9. AMQP returns DeclareOk-Sally in less than a second. This pushes DeclareOk-Sally on to ContinuationQueue's @q
10. wait_on_continuations pops DeclareOk-Bob from @q
11. queue calls "@name = queue_declare_ok.queue" changing it's own @name from "Sally" to "Bob"
12. channel calls register_queue with queue "Bob".
[...]

PR #566, opened 2018-08-28
Filter out mismatching (e.g. arriving after timeout) queue.declare-ok responses

michaelklishin:
When a belated response for a `queue.declare` arrives, it will be filtered out like in the Java client's continuation mechanism.

WIP because this can be refactored further and potentially extended to at least one more method (`exchange.declare`).

References #558.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -985,7 +985,8 @@ module Bunny
 
     # Declares a queue using queue.declare AMQP 0.9.1 method.
     #
-    # @param [String] name Queue name
+    # @param [String] name The name of the queue or an empty string to let RabbitMQ generate a name.
+    #                      Note that LF and CR characters will be stripped from the value.
     # @param [Hash] opts Queue properties
     #
     # @option opts [Boolean] durable (false)     Should information about this queue be persisted to disk so that it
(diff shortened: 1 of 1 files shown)


sbonebrake:
With this change I can no longe reproduce the issue in #558 which is great and i'm happy with this approach. However, I see the java code references an issue with this method.
https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/AMQChannel.java#L536
https://github.com/rabbitmq/rabbitmq-server/issues/710
I'm not certain how big of an issue that is, but something to consider.

michaelklishin:
@sbonebrake the server will only strip non-printable characters as it makes it next to impossible to delete such queues using CLI tools or management UI.

I am surprised that https://github.com/rabbitmq/rabbitmq-server/issues/710 was mentioned there at all as it's very much an edge case. @acogoluegnes do you think it's a major concern?

michaelklishin:
So this cannot be extended to `exchange.declare` without a protocol extension because `exchange.declare-ok` doesn't carry a name (exchanges cannot be server-named). Let's stop at this and see what other methods may be worth safeguarding the same way.

hosamaly:
This change breaks code that was using a `Symbol` for the queue name or the exchange name. If this is considered a good thing, it probably deserves to be mentioned on the changelog.

michaelklishin:
@hosamaly you are the first person to notice this in over three years. No docs that I recall use symbols for names. I consider this to be a "feature" that worked by accident, and a mitigation should be pretty straightforward.

PR #567, opened 2018-09-18
Improve recovery specs

baelter:
Some specs did not wait for connection recovery so it was random if the spec would pass or not.


michaelklishin:
Thank you!

PR #572, opened 2018-12-27
Removing effin dependency

lucianosousa:
Not required anymore since ruby 2.x has utf-8 as default encoding


lucianosousa:
@michaelklishin found this first than into `rabbitmq_http_api_client` and `amq-protocol`, but had to remove there for make it works fine! :D
Finally I did 🎉

PR #573, opened 2019-02-19
Added a network callback

ionutzp:
a new option for bunny to receive a Proc as a callback for when it re-establishes connections automatically

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -115,6 +115,7 @@ module Bunny
     # @option connection_string_or_opts [Integer] :read_timeout (30) TCP socket read timeout in seconds. If heartbeats are disabled this will be ignored.
     # @option connection_string_or_opts [Integer] :write_timeout (30) TCP socket write timeout in seconds.
     # @option connection_string_or_opts [Proc] :hosts_shuffle_strategy A Proc that reorders a list of host strings, defaults to Array#shuffle
+    # @option connection_string_or_opts [Proc] :network_recovery_callback A Proc that will be called when a network recovery is performed
     # @option connection_string_or_opts [Logger] :logger The logger.  If missing, one is created using :log_file and :log_level.
     # @option connection_string_or_opts [IO, String] :log_file The file or path to use when creating a logger.  Defaults to STDOUT.
     # @option connection_string_or_opts [IO, String] :logfile DEPRECATED: use :log_file instead.  The file or path to use when creating a logger.  Defaults to STDOUT.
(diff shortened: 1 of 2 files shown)


spec/higher_level_api/integration/connection_recovery_spec.rb
```diff
@@ -40,6 +49,19 @@
...
+    with_open(c_with_callback) do |c|
+      ch1 = c.create_channel
+      ch2 = c.create_channel
+      sleep 1.5
+      close_all_connections!
+      sleep 0.5
+      poll_until { channels.count == 2 }
+    end
+
+    expect(network_recovery_callback).to have_received(:call).exactly(1).times
```
michaelklishin: I have strong feelings about using mocks in data service clients such as Bunny (TL;DR: they are overused and mostly useless) but here it might be OK. I'm happy to update this test on my own.

michaelklishin:
I don't know how GitHub managed to lose two of my comments but here they are again.

Thank you. I feel the name you chose here could use some improvement. It suggests that it only covers the "network part" of connection recovery and doesn't hint at when the callback is meant to be executed.

Let's rename it to `connection_recovery_complete`.

lib/bunny/session.rb
```diff
@@ -797,12 +801,16 @@ def recover_channels
...
           ch.open
-
           ch.recover_from_network_failure
         end
       end
     end
 
+    # @private
+    def network_recovery_callback
+      @network_recovery_callback&.call
```
michaelklishin: Not sure what the trailing `&` does in this case but we can leave it with just `.call` as we do with all similar methods that accept callables.

ionutzp: hi, by default the callback will be `nil`, `&.` is the safe navigation operator to prevent nil class errors

spec/higher_level_api/integration/connection_recovery_spec.rb
ionutzp: here I just passed a double/mock/spy as a callback so i can see if it has been called, how would you approach this otherwise

lib/bunny/session.rb
michaelklishin: I have been using Ruby since 2005 and never knew about it. Can we please use a more obvious `… if callback` guard?

ionutzp:
is it ok now @michaelklishin

michaelklishin:
I've renamed the option, added a way to set the callback with a block and now reworking the test.

PR #564, opened 2018-08-22
Resolve a few warnings about uninitialized variables.

dacto:
consumer_work_pool.rb
  * During initialize, work pool is not running; set @running to false.

cruby/ssl_socket.rb, jruby/ssl_socket.rb
  * Add initialize method to init @__bunny_socket_eof_flag__ to false while
    still passing args, if any, to superclass

reader_loop.rb
  * Initialize flags used in the loop to sensible values given that:
      Not stopping or stopped (since we havent started), and don't know if
      network is down, yet.

session.rb
  * Move initialization of variables referenced by to_s before the call to
    init_default_logger, which calls to_s.
  * Initialize @transport, @heartbeat_sender, and @last_connection_error before
    they are referenced.
  * Removed code referencing @socket_configurator.
    The commit that added it: aa2642c8da2a294baa5d96dc4b11f56278283f90
    The commit that removed setting it: 7a1b5578e14acd69463909275b22a1293059a9ff
    The library doesn't sets it anywhere via instance_variable_set but users
    might've set it this way instead of using configure_socket..

transport.rb
  * Initialize @socket before it is referenced.

fixes #563

PR #581, opened 2019-08-01
Remove a couple of not initialized errors

wishdev:
This removes 2 more _instance variable @__bunny_socket_eof_flag__ not initialized_ warnings to add to the list from #564


michaelklishin:
Thank you. I decided to take a slightly different approach.

PR #590, opened 2020-02-04
Do not join @thread if that is the current Thread

### LINKED ISSUE #589: Forced connection close on Bunny with ConnectionPool can't recover
Hi,

I'm running into an issue using Bunny in a Rails project. I'm using the gem `connection_pool` to refrain from setting up a connection every time I want to send a message. The initializer for the pool looks as follows:

```
BUNNY_POOL = ConnectionPool.new(size: 5) do
  connection = Bunny.new(
    ENV['AMQP_URL'] || ::ExampleApp::Application.credentials.amqp_url,
  ).tap(&:start)

  connection.create_channel
end
```

I have a health check route which roughly does the following:

```
def health
  render json: {
    web: :ok,
    redis: redis_health,
    rabbitmq: rabbitmq_health,
  }
end
```

The `rabbitmq_health` method creates a reply queue and sends a ping message to it, nothing special here.

However, when I restart RabbitMQ or close the connection from the RabbitMQ management interface I get the following error:

```
W, [2020-02-02T12:06:09.585404 #52051]  WARN -- #<Bunny::Session:0x7f8f7f902f20 example_dev@localhost:5672, vhost=dev, addresses=[localhost:5672]>: Recovering from connection.close (CONNECTION_FORCED - Closed via management plugin)
#<Thread:0x00007f8f7fb8e0f0@/Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:35 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
  8: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:36:in `run_loop'
  7: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:36:in `loop'
  6: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:39:in `block in run_loop'
  5: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:92:in `run_once'
  4: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/session.rb:611:in `handle_frame'
  3: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/session.rb:862:in `clean_up_on_shutdown'
  2: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/session.rb:1029:in `maybe_shutdown_reader_loop'
  1: from /Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:119:in `join'
/Users/fuegas/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bunny-2.14.3/lib/bunny/reader_loop.rb:119:in `join': Target thread must not be current thread (ThreadError)
E, [2020-02-02T12:06:13.437710 #52051] ERROR -- #<Bunny::Session:0x7f8f7f902f20 example_dev@localhost:5672, vhost=dev, addresses=[localhost:5672]>: I/O error in the hearbeat sender: closed stream
```

It tries to recover from a closed connection but then tries to join a thread which it is not allowed to join resulting in an exception and breaking that connection. From that moment that connection will return:

```
Bunny::ConnectionClosedError: Trying to send frame through a closed connection. Frame is #\u003cAMQ::Protocol::MethodFrame:0x00007f8f89b64638 @payload=\"\\x002\\x00\\n\\x00\\x00\\x00\\x04\\x00\\x00\\x00\\x00\", @channel=1\u003e, method class is AMQ::Protocol::Queue::Declare"
```

To resolve the broken connection the webservice (Puma) needs to be restarted.

Do I have a flaw in the way I'm using Bunny? Is this caused by the usage of ConnectionPool?

(Gem) Versions:
- connection_pool: 2.2.2
- bunny: 2.14.3
- rabbitmq: 3.8.2 (Erlang 22.2.1)
- ruby: 2.6.5
- rails: 6.0.2.1
- puma: 4.3.1

fuegas:
If you're running multiple Bunny connections in a pooled manner, for example using the gem `connection_pool`, the threading model can be altered by the pooling gem to disable handling of exceptions.

For example, the gem `connection_pool` disables the handling when it creates an instance, and enables the handling when you're using a connection:

```ruby
  def with(options = {})
    Thread.handle_interrupt(Exception => :never) do
      conn = checkout(options)
      begin
        Thread.handle_interrupt(Exception => :immediate) do
          yield conn
        end
      ensure
        checkin
      end
    end
  end
```

When you've setup a connection and released it to the pool, it slumbers in a Thread of which the handling of exceptions is disabled. If you receive a (forced) disconnect or a NetworkError occurs, it triggers the recovery (when enabled) of Bunny. This recovery relies on a ShutdownSignal to be raised on the `reader_loop` thread which needs to be caught by the `clean_up_on_shutdown` in `Session`. Because the exception is not handled anymore, the code in `maybe_shutdown_reader_loop` in `Session` tries to join the `reader_loop` thread to force a re-raise of a thrown unhandled exception. That join is done on the `reader_loop` thread which is the same thread that started the recovery so we'll receive a ThreadError telling us "Target thread must not be current thread (ThreadError)".

To prevent the above from happening we can check if the thread we're trying to join is the same thread that is currently running and not join if it is the same thread.

This PR fixes #589.

### DIFF
--- a/lib/bunny/reader_loop.rb
+++ b/lib/bunny/reader_loop.rb
@@ -116,7 +116,7 @@ module Bunny
     end
 
     def join
-      @thread.join if @thread
+      @thread.join if @thread && @thread != Thread.current
     end
 
     def kill


michaelklishin:
Backported to `2.14.x-stable`. Please proceed with whatever you were doing before discovering #589, if no issues come up in a few days I will cut a release. Thanks again for contributing a really non-trivial usability improvement!

fuegas:
I'll patch our staging environment and run some rigorous tests on it. If anything pops up I'll leave a comment in #589 or #590.

fuegas:
@michaelklishin: just as an update, we've encountered no irregularities during testing. We did see correctly recovering connections.

michaelklishin:
OK, I'll cut a new release now. Thanks for following up.

michaelklishin:
[`2.14.4` is out](https://rubygems.org/gems/bunny)

PR #591, opened 2020-03-21
issue202_spec: don't connect to the internet

terceiro:
192.192.192.192 is a real server on the internet, and every time someone
runs the bunny test suite it will receive a connection attempt.

127.0.0.254 is a local address, and 1433 is the port for Microsoft SQL
Server. So, unless you are runinng the bunny test suite on a Microsoft
SQL server that happens to listen to all interfaces, the connection
attempt should always fail (and fail faster).


michaelklishin:
Thank you!

PR #592, opened 2020-03-21
exchange_recovery_spec: drop require_relative to lib

terceiro:
rspec already automatically adds lib to the $LOAD_PATH, and this hinders
initiatives such as Debian's practice of running the test suite against
the package installed system-wide (as opposed to against the source tree).


michaelklishin:
Thank you.

PR #598, opened 2020-08-13
Fixes #597 - Session error handler object which responds to #raise

### LINKED ISSUE #597: Bunny raises asynchronous exceptions on the thread that opened the connection
Bunny is raising asynchronous exceptions using `Thread#raise` in the thread that happened to call `Session#start`. This introduces a potentially dangerous and unpredictable behavior as the asynchronous exception (interrupt) can happen at any point in the thread lifetime, even in an `ensure` section, and is effectively impossible to defend against. 

The proposal is to provide means to specify a `session_error_handler` object that implements `#raise`, which if specified would be used instead of the origin thread, otherwise it would fall back to origin thread for backwards compatibility.

https://github.com/bbascarevic-tti/bunny/commit/4b282030bece8ecbfd250c5d39f381640066c03c

bbascarevic-tti:


michaelklishin:
Thank you!

Issue #599, opened 2020-09-10
Support custom connection_name


brerx:
Feature request:

rabbitmq-management has the ability to display a custom connection_name, if provided.

https://github.com/rabbitmq/rabbitmq-management/pull/172

As far as I can see, Bunny does not support this yet.

Can this be added, please?

michaelklishin:
This is open source software, so you are welcome to contribute the features that you need.

[Client-provided connection name](https://www.rabbitmq.com/api-guide.html#client-provided-names) is transferred in a client property, `connection_name`, which can be set with Bunny. A more convenient helper, Java and .NET client-style, can be added as an option and then [merged together with other client-provided and default client properties](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L201).

brerx:
Ok, thx for the hints. I will try to add it myself 👍

michaelklishin:
So, something like this:

``` ruby
conn = Bunny.new(connection_name: "app XYZ #{rand}")
conn.start
```

which would be equivalent to

``` ruby
conn = Bunny.new(client_properties: {connection_name: "app XYZ #{rand}"})
conn.start
```

hopewise:
@michaelklishin as this feature been released, we tried to use it but without luck

```
conn = Bunny.new(connection_name: "app XYZ #{rand}")
```


<img width="672" alt="Screenshot 2024-04-29 at 4 02 56 PM" src="https://github.com/ruby-amqp/bunny/assets/1061157/97c2a8b8-a642-4890-9ab6-43f40040d078">

brerx:
Hm, we have been using this heavily ever since. Works like a charm. So in general, it's working. I can't spot the problem with your code, either...

hopewise:
We use sneakers gem, it seems that it removes the connection name, the only way worked by setting it through sneakers gem its self https://github.com/jondot/sneakers/issues/410#issuecomment-780115384

michaelklishin:
@brerx I have previously provided evidence that this feature does work as expected in https://github.com/rabbitmq/rabbitmq-server/discussions/11109.

Without an executable way to reproduce or even Bunny (or Sneakers) and RabbitMQ versions mentioned, I conclude that the reporter has zero respect towards open source software maintainers' time => I personally disengage.

If someone figures out why Sneakers overrides this field and presents a convincing argument why it should not, I can merge PRs for Sneakers (but sadly do not have the permissions to produce new releases).

PR #600, opened 2020-09-11
Support connection name

brerx:
related to https://github.com/ruby-amqp/bunny/issues/599.

- adds connection_name as optional client_property
- adds unit test

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -201,7 +201,8 @@ module Bunny
 
       client_props         = opts[:properties] || opts[:client_properties] || {}
       @client_properties   = DEFAULT_CLIENT_PROPERTIES.merge(client_props)
-      @connection_name     = @client_properties.fetch(:connection_name, nil)
+                                                      .merge(connection_name: opts[:connection_name])
+      @connection_name     = @client_properties[:connection_name]
       @mechanism           = normalize_auth_mechanism(opts.fetch(:auth_mechanism, "PLAIN"))
       @credentials_encoder = credentials_encoder_for(@mechanism)
       @locale              = @opts.fetch(:locale, DEFAULT_LOCALE)
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -89,6 +89,7 @@ module Bunny
     # @return [Integer] Timeout for blocking protocol operations (queue.declare, queue.bind, etc), in milliseconds. Default is 15000.
     attr_reader :continuation_timeout
     attr_reader :network_recovery_interval
+    attr_reader :connection_name
     attr_accessor :socket_configurator
 
     # @param [String, Hash] connection_string_or_opts Connection string or a hash of connection options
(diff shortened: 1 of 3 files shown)


brerx:
@michaelklishin This is my first approach to solve this.

Without the addition in docker-entrypoint, the CLI tool proceeded before the actual users were created, which resulted in countless Auth errors during testing. So this should enable more people to contribute here.

I am eager to hear your feedback, especially whether the unit test in in the right place. I was not sure whether to put it rather into the integration/connection_spec...

docker/docker-entrypoint.sh
```diff
@@ -8,6 +8,7 @@ $server -detached
 
 echo "Waiting for RabbitMQ to finish startup..."
 
+sleep 10
```
michaelklishin, line 11: It makes no sense to sleep before `rabbitmqctl await_startup`. That command waits for node boot to complete. User creation is synchronous, so I don't see any need to wait after we set up the permissions but at least it would make some sense there.

michaelklishin (changes requested):

lib/bunny/session.rb
```diff
@@ -199,6 +201,7 @@ def initialize(connection_string_or_opts = ENV['RABBITMQ_URL'], optz = Hash.new)
 
       client_props         = opts[:properties] || opts[:client_properties] || {}
       @client_properties   = DEFAULT_CLIENT_PROPERTIES.merge(client_props)
+      @connection_name     = @client_properties.fetch(:connection_name, nil)
```
michaelklishin, line 204: `Hash#fetch` makes sense when you want a non-`nil` default. Should we use `Hash#[]` here?

spec/lower_level_api/integration/basic_consume_spec.rb
```diff
@@ -96,4 +96,10 @@
       ch.close
     end
   end
+
+  it "propagates the connection_name to the client_properties" do
```
michaelklishin, line 100: This belongs to connection tests. It has nothing to do with consumers.

michaelklishin (changes requested):

michaelklishin:
This does not actually populate the client property by merging it into `@client_properties`. So this makes it easy to access connection name in the app but not set it, which I thought was the goal?

docker/docker-entrypoint.sh
brerx: Ok, as an unrelated change, I'll remove this again. No problem.

michaelklishin: Yes, we can address this separately if needed. Admittedly I do not use this image for testing as I usually test Bunny against RabbitMQ nodes built from source or GA releases running locally.

brerx:
Ah now I think I know how you want to understand it: Instead of

```ruby
Bunny.new(properties: { connection_name: 'name' } }
```

which will merge automatically to the @client_properties, you thought of something like this:

```ruby
      @client_properties   = DEFAULT_CLIENT_PROPERTIES.merge(client_props).merge(connection_name: opts[:connection_name])

```

Am I right? I like this even better.

spec/lower_level_api/integration/basic_consume_spec.rb
brerx: done

michaelklishin:
Correct, it should be easy to set a connection name, reading it is primarily useful for tests and, perhaps, logging. 

`Bunny.new(properties: { connection_name: 'name' } }` would still work. I'm not sure whether `connection_name` the option or client_properties should take precedence, your approach seems right.

brerx:
Ok, I changed it to the easier usecase. What do you think?

brerx:
Thx  for merging this so fastly. Will you make a new release soon?

michaelklishin:
[2.17.0](https://rubygems.org/gems/bunny/versions/2.17.0) is up on rubygems.org.

PR #604, opened 2020-12-07
SizedQueue is used in the consumer work pool to avoid memory leaks.

### LINKED ISSUE #603: huge memory leak when consumer pool can't handle the load
If the number of incoming messages is greater than the consumer is handling, the memory usage increases infinitely over time.
The simplest ruby ​​script can use a huge amount of memory
```
require 'bunny'

Process.setproctitle("test_consumer")

conn = Bunny.new(
# bunny conf
)
conn.start
channel = conn.create_channel

exchange = channel.send(:fanout, 'test')
queue = channel.queue('test_queue', exclusive: false)
queue.bind(exchange)

queue.subscribe do |*args|
  puts args.inspect[0..100]
  sleep 0.01 # mock message handling
end

Thread.list.reject { |t| t == Thread.current }.each(&:join)
```

and this memory is not released even after reducing the load

leksito:
Fixes #603

### DIFF
--- a/lib/bunny/consumer_work_pool.rb
+++ b/lib/bunny/consumer_work_pool.rb
@@ -17,13 +17,13 @@ module Bunny
     attr_reader :size
     attr_reader :abort_on_exception
 
-    def initialize(size = 1, abort_on_exception = false, shutdown_timeout = 60)
+    def initialize(size = 1, abort_on_exception = false, shutdown_timeout = 60, pool_queue_limit=10000)
       @size  = size
       @abort_on_exception = abort_on_exception
       @shutdown_timeout = shutdown_timeout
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you for taking the time to contribute.

This won't address the fundamental issue with consumer overload. A sensible [prefetch value on the channel](https://www.rabbitmq.com/confirms.html) will since that will control the delivery rate. With this change the in-flight deliveries will accumulate elsewhere.

Picking a good buffer size in cases such as this is effectively impossible for a library like Bunny since it has no context of any kind about application-level metrics and what ranges are considered to be healthy. So at the very least this has to be a configurable option but I doubt we need it given that that's exactly what prefetch and manual acknowledgements accomplish already.

PR #605, opened 2020-12-17
fix automatically_recover config option

KaifasKain:
If option **automatically_recover: false** and **automatic_recovery** not set

`opts[:automatically_recover] || opts[:automatic_recovery] #=> nil
`
Other cases
```
false || nil #=> nil
nil || false #=> false

false | nil #=> false
nil | false #=> false
```
automatically_recover? return nil
https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L456

```
# @return [Boolean] true if this connection has automatic recovery from network failure enabled
def automatically_recover?
  @automatically_recover
end
```

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -172,7 +172,7 @@ module Bunny
       @automatically_recover = if opts[:automatically_recover].nil? && opts[:automatic_recovery].nil?
                                  true
                                else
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
Thank you!

PR #617, opened 2021-06-14
Don't reconfirm already acked messages when multiple is true

### LINKED ISSUE #615: Redeliver publisher confirms acks again and again from 0 till current tag
I wrote some benchmarks and met some strage behaviour:
In confirm_select mode sometimes callback called with already acked delivery tag. I start to investigate and find this code:
https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/channel.rb#L1792

When rabbitmq sends me an confirm with `multiple` equals to true `confirmed_range_start` setted to `@delivery_tag_offset + 1`. But @delivery_tag_offset is always 0 without network recoveries!  

So I have such results:
Send 16000 very small messages
ConfirmSelect callback called 119567 times - each delivery_tag was acked ~ 9 times. 

Which workaround can I use? The main fix I see to pass tag|multiple|nack triplet to user in callback without changes, avoiding useless work to iterate through tags from beginning over and over

kinnalru:
In [current implementation ](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/channel.rb#L1792) library acks all tags starting from tag **1** (plus channel recovery offset) every time it gets `multiple` flag regardless of previously acknowledged tags. So when `multiple` flag occured ex. at delivery_tag 1000 (after 990 aked), we receive `confirm_select callback` 1000 times. And we receive this call callback again and again (hudren thousands times) while processing just 16k messages.

After some research of other language implementation ([java](https://github.com/rabbitmq/rabbitmq-java-client/blob/394d8968d8babe64c15357d53debfe3aacfbbf24/src/main/java/com/rabbitmq/client/impl/ChannelN.java#L1604) and [python](https://github.com/pika/pika/blob/1baf475547351c4f37762c5534de9ac49affed3d/pika/adapters/twisted_connection.py#L715)) we made a little fix to resolve this issue. [RabbitMQ protocol extension](https://www.rabbitmq.com/confirms.html#publisher-confirms) strgongly enforces sequenced delivery_tag generation on publishing. There is `@unconfirmed_set` filled with not-yet-acked tags (previously acked tags already removed from it) so we just need to ack tags from first (smallest) unconfirmed position till currentlry delivered with `multiple` flag to fit RabbitMQ publisher confirms protocol extension. 

There is additional test case to test and compare old and new behaviour. It is not for enclusion in resulting PR - just for experiments.  You can change messages count to higher value when **difference increased drastically**.

Fixes #615


michaelklishin:
Thank you for the thorough PR. Note that you can add benchmarks under `benchmarks`, e.g. we can keep the one that tests the proposed implementation there. Or we can drop it as it's served its purpose.

My test runs on a 4-year old CPU suggest that once the number of messages published goes above 20K or so, the new implementation begins significantly outperforming. In general, this implementation should be acceptable
for most users even if `Set#min` is not optimised to be a constant time operation.

So I'd be happy to merge once you remove the temporary tests (or move one of them to `./benchmarks`). Thanks again for
putting in the time and effort 👏

yurusov:
I deleted those specs, they've been created only for this PR, to show difference in performance. Thanks for your time and quick response!
P.S. Sorry for extra pr I made by mistake.

michaelklishin (approved):

PR #621, opened 2021-06-25
Remove test_files from gemspec

alexeyschepin:
Seem to `s.test_files` is already redundant and has been deleted from many other gems rubygems/rubygems#735

By removing test files from the gem, it becomes three times lighter (from 159Kb to 54Kb)
It will also stop annoying image screening tools for containing private keys from `spec/tls` folder.


michaelklishin:
Test certificate and private key pairs are already excluded as of a0ec20dd351954b36880c74e33b515afbdc8257d.

michaelklishin (approved):

michaelklishin:
Thanks. I will tag and push a `2.19.1` release in the next day or two.

PR #622, opened 2021-08-17
Update documentation to match used opts

KonradChlupka:
The Session.new doc specifies tls_version as one of the options but this
doesn't have any effect. In `class Transport`, the actually used options
are `ssl_version = opts[:tls_protocol] || opts[:ssl_version]`.


michaelklishin:
Thank you!

PR #625, opened 2021-11-19
Fix #to_inspect and #to_s in Bunny::Channel and Bunny::Consumer

lacrosse:
These caused some confusion while trying to parse a huge debug output.


michaelklishin (approved):

michaelklishin:
Thank you!

Issue #628, opened 2022-03-11
TLS 1.3 support


chevinbrown:
Bunny version: 2.0.19
RabbitMQ version: 3.9.13
Ruby: 2.7

Based on ruby's openssl lib, it looks like you must use `min_ssl_version` to use 1.3: https://github.com/ruby/openssl/issues/473#issuecomment-964252959

Using the option `TLSv1_3`, 
```
Bunny.new("amqps://", ... tls_protocol: :TLSv1_3)
...
#=> ArgumentError: unknown SSL method `TLSv1_3'
```
It doesn't look like there is the ability to pass args through to openssl?
Is 1.3 support on the horizon?

michaelklishin:
@chevinbrown this is open source software, so things "get on the horizon" when someone motivated contributes them or at least clearly explains what has to be done by others.

Bunny allows you to [configure TLS context any way you can in Ruby code](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/transport.rb#L126).

Bunny could use `OpenSSL::SSL::SSLContext#min_version=` internally but that method is [no available in Ruby versions older than `2.5.0`](https://docs.ruby-lang.org/en/2.4.0/OpenSSL/SSL/SSLContext.html). I see Bunny README already states 2.5 is the minimum version supported, which I doubt is actually the case but very convenient.

michaelklishin:
As of #629, Bunny uses the modern methods described in https://github.com/ruby/openssl/issues/473#issuecomment-964252959.

Since [TLS 1.3 is an all-or-nothing proposition](https://rabbitmq.com/ssl.html#tls1.3), I'm not sure what to do about
the test suite. As soon as I get a TLS 1.3 example running without any alerts from the server, this can be closed
with a new release. I cannot promise when that might happen but with #629, anyone can try it. We've had TLS 1.3
documented in RabbitMQ docs for months now.

PR #629, opened 2022-03-22
Use modern SSLContext API

michaelklishin:
It's already possible to post-configure TLS context using
`Bunny::Connection#transport` and `Bunny::Transport#configure_tls_context`
but since we already require Ruby 2.5, using a more modern
API makes sense. It also allows the user to set TLS version to 1.3 [1].

Note that TLS 1.3 support is an all-or-nothing proposition
because there are no shared cipher suites with earlier versions,
by design.

References #628.

1. https://rabbitmq.com/ssl.html#tls1.3

### DIFF
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -25,6 +25,24 @@ module Bunny
     DEFAULT_READ_TIMEOUT  = 30.0
     DEFAULT_WRITE_TIMEOUT = 30.0
 
+    # mimics METHODS_MAP in ssl.rb but also lists TLS 1.3
+    # and string constants
+    TLS_VERSION_ALIASES = {
+      TLSv1: OpenSSL::SSL::TLS1_VERSION,
+      TLSv1_1: OpenSSL::SSL::TLS1_1_VERSION,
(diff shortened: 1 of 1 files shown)


michaelklishin:
With a RabbitMQ 3.9 node on Erlang 24.3 with the following `rabbitmq.conf`:

``` ini
#
# TLS
#

listeners.ssl.default  = 5671
ssl_options.cacertfile = /path/to/tls-gen.git/basic/result/ca_certificate.pem
ssl_options.certfile   = /path/to/tls-gen.git/basic/result/server_certificate.pem
ssl_options.keyfile    = /path/to/tls-gen.git/basic/result/server_key.pem
```

I could not make one Bunny TLS test example modified to use TLS 1.3. The node kept logging

```
[notice] <0.1324.0> TLS server: In state start at tls_handshake_1_3.erl:2964 generated SERVER ALERT: Fatal - Illegal Parameter
```

and it's not obvious what that "illegal parameter" is without a traffic dump and extra digging I cannot do right now.

But all existing TLS 1.2 tests pass, so that's a step in the right direction.

PR #635, opened 2022-07-13
Avoid race between multi-message ack and publish

romuloceccon:
PR #617 introduced an optimization for multi-message acks but failed to protect @unconfirmed_set. In edge cases #basic_publish would attempt to modify it during iteration with Enumerable#min:

    <ruby>/lib/ruby/3.1.0/set.rb:522:in `add': can't add a new key into hash during iteration (RuntimeError)
    from <bunny>/lib/bunny/channel.rb:555:in `block in basic_publish'
    from <bunny>/lib/bunny/channel.rb:554:in `synchronize'
    from <bunny>/lib/bunny/channel.rb:554:in `basic_publish'
    from <bunny>/lib/bunny/exchange.rb:141:in `publish'

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -1788,12 +1788,12 @@ module Bunny
 
     # @private
     def handle_ack_or_nack(delivery_tag_before_offset, multiple, nack)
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you!

PR #639, opened 2022-09-14
Fix delivery tag offset calculation after network failure recovery

remisauvat:
This happens when using confirm_select and receiving ack with multiple = true

The range of confirmed tag was not calculated properly after a network recovery and the update of the delivery_tag_offset value.

### DIFF
--- a/lib/bunny/channel.rb
+++ b/lib/bunny/channel.rb
@@ -143,6 +143,10 @@ module Bunny
     attr_reader :work_pool
     # @return [Integer] Next publisher confirmations sequence index
     attr_reader :next_publish_seq_no
+    # @return [Integer] Offset for the confirmations sequence index.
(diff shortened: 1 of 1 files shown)


VasiliCekaskin:
I am not familiar with bunny, but for me it is hard to understand what the problem is and why it occurs by looking at this PR. I am not familiar with the library at all to be honest, but by reading this PR and your initial comment it is hard for me to find out without reading & understanding a lot of the code.

It might be just me, but I think it would be a good practice to describe the problem that you are trying to fix without assuming that everyone knows what you are talking about in this issue. That way we can learn about the library without knowing every aspect of it.

michaelklishin:
I wish there was a way to cover this with tests, even unit ones would be a significant improvement.

PR #649, opened 2023-01-10
Fix non bang merge

rene-muehlboeck:
A quorum queue can't be created because the `x-queue-type` argument isn't merged into `@options[:arguments]`


michaelklishin:
But the merged map is returned by the if expression. I will try to reproduce but do not like mutating methods like this where they can be avoided.

michaelklishin:
A slightly different solution is out in `2.20.2`. Thank you.

rene-muehlboeck:
Thank you for the quick solution!

PR #657, opened 2023-02-17
Correct base OpenSSL exception class in modern Rubies

michaelklishin:
See #656.


michaelklishin:
@womblep does this help your case?

womblep:
@michaelklishin no I dont think it does. I have added to the discussion. I think the exact Exception is not relevant but it is where they are trapped

PR #658, opened 2023-02-21
issue656 - change recoverable_network_failure? 

womblep:
Alter recoverable_network_failure? to check against a definable array of exceptions and modify recover_from_network_failure to handle the exception logic differently

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -71,7 +71,7 @@ module Bunny
     # Default reconnection interval for TCP connection failures
     DEFAULT_NETWORK_RECOVERY_INTERVAL = 5.0
 
-    DEFAULT_RECOVERABLE_EXCEPTIONS = [TCPConnectionFailedForAllHosts, TCPConnectionFailed, AMQ::Protocol::EmptyResponseError, SystemCallError, Timeout::Error, Bunny::ConnectionLevelException, Bunny::ConnectionClosedError]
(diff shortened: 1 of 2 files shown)


michaelklishin (changes requested):
This looks good but I get two spec failures on Ruby 3.2 (macOS on ARM64):

```
Failures:

  1) Bunny::Session when retry attempts have been exhausted closes the session
     Failure/Error: expect(session.closed?).to be true

       expected true
            got false
     # ./spec/issues/issue549_spec.rb:20:in `block (3 levels) in <top (required)>'

  2) Bunny::Session when retry attempts have been exhausted stops the reader loop
     Failure/Error: expect(reader_loop.stopping?).to be true

       expected true
            got false
     # ./spec/issues/issue549_spec.rb:27:in `block (3 levels) in <top (required)>'

Finished in 42.48 seconds (files took 0.2137 seconds to load)
26 examples, 2 failures

Failed examples:

rspec ./spec/issues/issue549_spec.rb:17 # Bunny::Session when retry attempts have been exhausted closes the session
rspec ./spec/issues/issue549_spec.rb:23 # Bunny::Session when retry attempts have been exhausted stops the reader loop
```

michaelklishin:
This is a test inter-dependency. When I run just one spec file, it passes 100 times out of 100.

When I run

``` ruby
rspec spec/issues/
```

then I get these two failures.

womblep:
This is a test case for a different issue. Does it happen before the change too?

womblep:
I picked out all the errors that I could find that were defined and added them to Session->DEFAULT_RECOVERABLE_EXCEPTIONS
StandardError was not one of them, should it be?
If I change the test to:
`      session = create_session`
`      session.recoverable_exceptions << StandardError`
`      session.handle_network_failure(StandardError.new)`
then it passes, suggesting that StandardError should be trapped

michaelklishin:
It's a good question since `StandardError` is almost a catch-all in Ruby. But we can try this.

michaelklishin (approved):

PR #659, opened 2023-03-23
Ruby 3.2 compatibility for 0.7.x stable

skaes:
We're still stuck on this old branch, but want to use ruby 3.2.

Are you OK with releasing a new 0.7.x version?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,7 @@ group :development do
 end
 
 group :test do
-  gem "rspec", "~> 3.6.0"
+  gem "rspec"
 end
 
 gemspec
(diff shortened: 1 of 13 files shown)


michaelklishin:
I would not have any way of testing it but if you'd like me to cut a new 0.7.x patch release, I can try.

skaes:
Thx a lot for supporting this. We are are already running [with a similar patch](https://github.com/xing/beetle/commit/e8eec47273f76aa593ce17f692866848b0a79066) in production. I'm confident it works.

skaes:
Just discovered that I'm also an owner of the gem on rubygems.org. I'll publish the new gem version myself.

PR #660, opened 2023-04-19
Stop logging heartbeat at INFO level

ccdredzik:
When using bunny in conjunction with sneakers, we are being spammed with the message that is being removed here. There is another message generated into DEBUG line above. Should we just leave it there?


ccdredzik:
@michaelklishin can I ask for a review of this one?

michaelklishin:
This message is logged once when connection is started, not every time a heartbeat is sent or received (those are logged, or at least the frames are, at `debug` level).

If you see a lot of these messages, you do not use connections [the way they were designed to be used](https://rabbitmq.com/connections.html#high-connection-churn).

Issue #665, opened 2023-06-09
Missing client notification when recovery fails (after specified recovery attempts)


Schmitze333:
Bunny: latest
RabbitMQ: 3.11.16
Ruby: 3.2.2

## Overview

The Bunny gem conveniently allows to specify a `recovery_attempts` option limiting the number of connection recovery retries.
However, when running into the situation that all recovery attempts are used up and a recovery was not successful, the application has no chance to be informed about that state.

## Example to reproduce

```ruby
require "bunny"

error = false

connection = Bunny.new(recovery_attempts: 2)
connection.start

channel = connection.create_channel
channel.on_uncaught_exception { error = true }
queue = channel.queue("hello")

consumer_thread =
  Thread.new do
    consumer = Bunny::Consumer.new(channel, queue)
    consumer.on_delivery { |_di, _m, payload| puts payload }
    queue.subscribe_with(consumer)
  end

consumer_thread.abort_on_exception = true

begin
  sleep 1 while !error
rescue Interrupt => _
  connection.close

  exit(0)
rescue => e
  pp e

  exit(1)
end
```

_Logs when interrupting the connection to RMQ:_

```
W, [2023-06-09T14:28:37.811469 #6429]  WARN -- #<Bunny::Session:0x11d0 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Recovering from connection.close (CONNECTION_FORCED - broker forced connection closure with reason 'shutdown')
W, [2023-06-09T14:28:37.811907 #6429]  WARN -- #<Bunny::Session:0x11d0 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Will recover from a network failure (2 out of 2 left)...
W, [2023-06-09T14:28:47.813447 #6429]  WARN -- #<Bunny::Session:0x11d0 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Retrying connection on next host in line: 127.0.0.1:5672
W, [2023-06-09T14:28:47.814373 #6429]  WARN -- #<Bunny::Session:0x11d0 guest@127.0.0.1:5672, vhost=/, addresses=[127.0.0.1:5672]>: Could not establish TCP connection to 127.0.0.1:5672: Connection refused - connect(2) for 127.0.0.1:5672
[...]

PR #666, opened 2023-06-09
Introduce recovery_failed callback

Schmitze333:
It’s possible to specify the recovery_attempts option on the Bunny::Session. 
Hitting the recovery attempts limit without a successful recovery leads to a cleanup of connection, etc.
However, the client is not informed about this state and has no chance to react upon it.

This PR introduces a callback which is executed in exactly the above described situation enabling client code to react to a failed recovery process.

Relates to: https://github.com/ruby-amqp/bunny/issues/665

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -128,6 +128,7 @@ module Bunny
     # @option connection_string_or_opts [Integer] :reset_recovery_attempts_after_reconnection (true) Should recovery attempt counter be reset after successful reconnection? When set to false, the attempt counter will last through the entire lifetime of the connection object.
     # @option connection_string_or_opts [Proc] :recovery_attempt_started (nil) Will be called before every connection recovery attempt
     # @option connection_string_or_opts [Proc] :recovery_completed (nil) Will be called after successful connection recovery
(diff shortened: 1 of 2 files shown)


michaelklishin (changes requested):
Thank you for taking the time co contribute.

"recovery_failed" to me suggests that this callback will be used for *every* unsuccessful recovery attempts. If that's not the case, we should rename it "recovery_attemtps_exhausted" or something that clearly suggests that this callable will be invoked only once
for a series for unsuccessful recovery attempts.

Schmitze333:
Yeah, I agree. That name is much better!
I'll make the necessary changes.

michaelklishin (approved):

michaelklishin:
@Schmitze333 thank you for contributing to Bunny!

Schmitze333:
I did not provide a Changelog entry. Will that happen on the release or shall I add one?

michaelklishin:
I update change log before every release if necessary. Don't worry about it.

michaelklishin:
[`2.22.0` is up](https://rubygems.org/gems/bunny) on rubygems.org.

PR #670, opened 2023-08-21
Fix exception on heartbeat sender

mRudzki:
Why?
This was introduced in:
https://github.com/ruby-amqp/bunny/commit/55d57e38a31ae787d8693fe6b6acbe0ccd515653
But due to change in:
https://github.com/ruby-amqp/bunny/commit/c85c24bc1210e028276d32994e9753b474d6f674
We lost logging of standard errors.
It is


michaelklishin:
You are correct, the intent was not to avoid handling `StandardError` decedents.

PR #693, opened 2024-11-14
Update docker-compose setup for tests

sorcix:
The Bintray repository used in the docker file no longer exists.

This commit updates the Dockerfile to use a recent version of Ubuntu and then follows the guidelines given by the RabbitMQ project for Debian and Ubuntu based distro's.

https://www.rabbitmq.com/docs/install-debian


michaelklishin:
Thank you!

PR #694, opened 2024-11-14
Fix queue_exists? for exclusive queues

sorcix:
Queues with the exclusive flag set can't be redeclared, causing a ResourceLocked exception. This solution assumes that if something is locked it also exists.

### DIFF
--- a/lib/bunny/session.rb
+++ b/lib/bunny/session.rb
@@ -516,6 +516,8 @@ module Bunny
       begin
         ch.queue(name, :passive => true)
         true
+      rescue Bunny::ResourceLocked => _
+        true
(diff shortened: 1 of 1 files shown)


michaelklishin:
For exclusive queues, this is a reasonable assumption. Thank you.

michaelklishin (approved):

michaelklishin:
@sorcix do you need a new release very soon or consuming Bunny from git would be acceptable for you?

sorcix:
We temporarily fixed it by placing an additional `rescue Bunny::ResourceLocked` around the single call to `queue_exists?` in our code, so no need for a new release.

PR #707, opened 2025-02-23
Unify ci and local testing

matias-martini:
WIP


michaelklishin:
@matias-martini we already use the same image. I'm not going to test against RabbitMQ 3.9, it has been [out of support for years](https://www.rabbitmq.com/release-information).

matias-martini:
Hi @michaelklishin,

I was meant to leave a description on this PR, but I had an emergency and had to leave it as WIP—sorry about that! 🙇‍♂️

Regarding dropping RabbitMQ 3.9 from CI, I completely understand. It was just a quick test to see if we were getting the same results as before the change and to explore how we could use it.

The main reason for this PR is that while working with the gem locally, I ran into issues with the Dockerfile. It wasn’t working out of the box, and I started wondering whether building a custom RabbitMQ image was necessary. Using the official RabbitMQ image seemed like a more maintainable approach.

The CI changes here are not essential and can be removed if needed. That said, I wonder if there's value in ensuring the local and CI setups are identical?

Another thing I’m unsure about is defining the RabbitMQ topology via definitions.json. While it simplifies containerized setups, it might make using a non-Dockerized RabbitMQ instance less straightforward. If needed, I could reintroduce a shared script to set up the topology while keeping the same logic.

That said, this approach reduces the amount of code we need to maintain while ensuring that the local environment matches exactly what CI will use to test the code. If you feel this doesn’t improve maintainability or make testing easier, I’m happy to drop the PR—just let me know! Appreciate your time and feedback. 🙏

matias-martini:
Following up on my previous comment, here is a little more information about the issue I mentioned when trying to build the RabbitMQ image.

The build fails at this step:

```sh
RUN apt-get install -y rabbitmq-server
```

with the following message:

``` sh
The following packages have unmet dependencies:
rabbitmq-server : Depends: erlang-base (>= 1:26.0) but 1:25.3.2.8+dfsg-1ubuntu4 is to be installed
                  or erlang-base-hipe (>= 1:26.0) but it is not installable
                  or esl-erlang (>= 1:26.0) but it is not installable
```

It shows that the RabbitMQ package requires Erlang 26, but the package [sources for Ubuntu 24.04 (noble) are only providing Erlang 25](https://www.ubuntuupdates.org/package/core/noble/main/base/erlang), causing a dependency conflict.

Maybe if we use the official RabbitMQ image, we could avoid these kinds of problems altogether, since it is already built with the correct Erlang version and dependencies (and not maintenance from the bunny team is needed)?

michaelklishin:
@matias-martini our team provides [cutting edge Erlang packages](https://www.rabbitmq.com/docs/install-debian) but we already use the community Docker image [on CI](https://github.com/ruby-amqp/bunny/blob/main/.github/workflows/ci.yml).

matias-martini:
@michaelklishin thanks for sharing! Just to clarify, would you prefer to keep maintaining the existing Dockerfile in this repo (fixing the current issue of dependencies) rather than switching to the official image **for local development**? If so, I’ll go ahead and drop the PR. Appreciate your time and feedback! 🙇🏻‍♂️

michaelklishin:
@matias-martini no, using the same image for local development also sounds good.

matias-martini:
@michaelklishin great! In that case, before making the necessary modifications to the PR, I’d love to get your thoughts on whether you prefer using a definitions file (definitions.json) loaded via rabbitmq.conf as done in this PR or a script to initialize the RabbitMQ topology.

I find the definitions file approach cleaner, but it ties local development to Docker Compose, which helps standardize environments. A script, on the other hand, allows setting up the topology on local RabbitMQ instances outside of Docker, which might be preferable for this project.

Would love to hear your preference on this!

michaelklishin:
A definitions file with password hashes would require a comment somewhere.

Plus there are things you cannot configure via the definitions file, such as the metric emission interval that's greatly reduced from the default 5s by most client library setup scripts.

Issue #700, opened 2025-02-20
Consider removing versioned delivery tags


michaelklishin:
From their inception in Jan 2013, they have been one element in Bunny's approach to connection
recovery that I've always had doubted.

Bunny's connection recovery approach was [ported](https://www.rabbitmq.com/client-libraries/java-api-guide#recovery) to both Java and .NET clients, and has stood the test of time. However, version tagging specifically was not adopted by most Bunny-inspired AMQP 0-9-1 clients.

All other parts of Bunny's connection recovery are easy to reason about and explain, as years of experience with the [RabbitMQ Java](https://www.rabbitmq.com/client-libraries/java-api-guide#recovery) and .NET clients demonstrate. That's not the case with versioned delivery tags, which are not a protocol concept.

Removing versioned delivery tags is one of the low hanging simplification fruits.

Issue #703, opened 2025-02-21
Consider folding Bunny::Transport into Bunny::Session


michaelklishin:
`Bunny::Transport` was originally introduced to separate `Bunny::Session`, a part of the original 2009 design, from the 2 socket implementations (MRI, JRuby) x two socket types ("plain" TCP, TLS).

As of #701 #702, this layer layer is likely unnecessary and `Bunny::Transport` can be merged with `Bunny::Session`.

## Backwards Compatibility Considerations

```ruby
c = Bunny.new
c.transport.configure_socket |socket|
end
```

can be reimplemented by making `Bunny::Session#transport` return `self`.

michaelklishin:
After spending an hour reviewing the responsibility split between `Bunny::Session` and `Bunny::Transport`, and how connection recovery is performed, I conclude that this is not worth the effort.

The separation of a `Transport` is reasonable and simplifies `Bunny::Session`.

What really would improve recovery safety w.r.t. #699 #709 is #704, not folding `Bunny::Transport` into `Bunny::Session`.

That said, after #704 we can choose to revisit this idea one more time.

Issue #701, opened 2025-02-20
Drop JRuby support


michaelklishin:
We have been recommending [March Hare](https://github.com/ruby-amqp/march_hare) for years.

PR #709, opened 2025-04-16
Pin delivery tags to the originating transport/socket

KlimSemikin:
Hello. This PR updates DeliveryInfo implementation to pin DeliveryTag to the
originating `Transport` instance. This ensures that we do not send delivery tags
on a different connection, causing PRECONDITION_FAILED error and removal of the
channel/consumers/etc from `Session` state. It also fixes various instances of
`to_i` being applied before `guarding_against_stale_delivery_tags`, making the
latter no-op.

See also https://github.com/ruby-amqp/bunny/issues/700


michaelklishin:
This is not at all what I had in mind in #700. The intent of #700 was to delete everything related to versioned tags, that's it.

`DeliveryInfo` should not be aware of the transport used.

I suspect that this is coming from the same team that https://github.com/ruby-amqp/bunny/discussions/699. So the solution to a "fundamentally broken design" is to entangle more objects that have little to do with each other?

michaelklishin:
I fail to see why several `Bunny::Channel` methods that have to do with consumer acknowledgements had to be axed.

Sorry, just like #699, this is an example of how not to contribute to open source. This PR changes too many things at once, the solution lumps together entirely unrelated things, several perfectly reasonable methods that've existed since 2010-2011 were removed seemingly for no reason.

There were no discussions as to whether I'd even consider a change like this. But of course there is an upvote fest 👍🚀 from a collegue.

Yeah, back to the drawing board.

michaelklishin:
#700 #701 are both done.

#703 is up next.

PR #716, opened 2025-05-19
Handle EBADF together with other SystemCallErrors

michaelklishin:
References https://github.com/ruby-amqp/bunny/discussions/715.

This combines both handlers into one that

 * Sets a value that will terminate the next loop
 * Tells the `Bunny::Session` parent to begin connection recovery


michaelklishin:
Backported to `2.x`.

PR #735, opened 2026-04-23
Ruby 3.0 modernisation, code quality improvements, and gem cleanup

eglitobias:
## Motivation

These messaging gems are critical dependencies in our production app, so I've been investing time in modernising the RabbitMQ Ruby ecosystem (hutch, amq-protocol) and want to contribute similar cleanup to Bunny. This PR removes accumulated legacy code and tightens things up for the Ruby 3.0+ baseline Bunny already requires.

## Changes

**Ruby 3.0 modernisation**
- Remove Ruby 2.0/2.1 compatibility guards (`defined? Encoding`, `respond_to?(:report_on_exception)`, encoding magic comments, `require "thread"`, `require "monitor"`)
- Sweep the remaining `:key =>` hash rocket syntax in `lib/` over to `key:`

**Bug fix**
- Several queue helpers (`quorum_queue`, `stream`, `durable_queue`, `delayed_queue`, `jms_queue`, `queue`) used `throw ArgumentError.new(...)` to signal validation errors. With no matching `catch`, this raises `UncaughtThrowError` instead of `ArgumentError`. Switched to `raise ArgumentError, "..."` and updated the specs that asserted the wrong error class

**Code quality**
- Deduplicate `@channels[ch_number]` hash lookup in `handle_frameset` — single lookup per frame instead of two
- Replace `respond_to?(:name)` duck-type checks with `is_a?(Bunny::Exchange)` / `is_a?(Bunny::Queue)` in the publish, bind, and consumer hot paths

**Gem / Gemfile cleanup**
- Remove redundant shebang and encoding comments from gemspec and Gemfile
- Remove duplicate vendor block from Gemfile
- Pin `amq-protocol` to `~> 2.7` to match the runtime dependency

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,46 +1,7 @@
-# encoding: utf-8
-
 source "https://rubygems.org"
-
-# Use local clones if possible.
-# If you want to use your local copy, just symlink it to vendor.
-# See http://blog.101ideas.cz/posts/custom-gems-in-gemfile.html
-extend Module.new {
-  def gem(name, *args)
-    options = args.last.is_a?(Hash) ? args.last : Hash.new
-
-    local_path = File.expand_path("../vendor/#{name}", __FILE__)
-    if File.exist?(local_path)
-      super name, options.merge(path: local_path).
-        delete_if { |key, _| [:git, :branch].include?(key) }
-    else
-      super name, *args
-    end
-  end
-}
-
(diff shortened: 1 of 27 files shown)


michaelklishin:
The failures are connection recovery flakes. I will get to them eventually.

Issue #737, opened 2026-06-02
subscribe_with/basic_consume_with consumers lose identity and callbacks after network recovery


jollopre:
When a consumer is registered via `Queue#subscribe_with` or `Channel#basic_consume_with`, automatic network recovery silently replaces the original
  `Bunny::Consumer` instance in `Channel#consumers` with a freshly constructed one.

  ## Affected versions

  Bunny 3.0.0, 3.1.0 (regression introduced in 3.0.0)

  ## Symptoms

  After a network failure and recovery:

   * `channel.consumers[tag].equal?(original_consumer)` returns `false` for consumers registered via `subscribe_with` or `basic_consume_with` — the original instance
   has been replaced
   * Any `on_cancellation` callbacks attached to the original consumer before recovery never fire on broker-initiated cancel

  ## Root cause

  In Bunny 2.x, `Consumer#recover_from_network_failure` called `@channel.basic_consume_with(self)`, preserving the original instance. In 3.0.0 that method was
  removed when topology recovery moved to `Session` + `TopologyRegistry`. The new recovery path in `Session#recover_consumer` always calls block-form
  `basic_consume`, which goes through `add_consumer` and creates a fresh vanilla `Bunny::Consumer` — regardless of how the consumer was originally registered.

  ## Reproduction

  ```ruby
  conn = Bunny.new(recover_from_connection_close: true)
  conn.start

  ch = conn.create_channel
  q  = ch.queue("", exclusive: true)

  consumer = Bunny::Consumer.new(ch, q)
  consumer.on_cancellation { puts "cancelled!" }
  q.subscribe_with(consumer)

  original_tag = consumer.consumer_tag

  # close the connection via the RabbitMQ management API and wait for recovery

  puts ch.consumers[original_tag].equal?(consumer) # => false (expected: true)
  # cancellation callback will never fire

  Expected behaviour

  - channel.consumers[tag].equal?(original_consumer) returns true after recovery
  - on_cancellation callbacks registered before recovery continue to fire

michaelklishin:
@jollopre thank you for the detailed analysis and the PR.

michaelklishin:
@jollopre we have enough changes to ship as `3.2.0`, it will happen in the next few days. Thanks again.

jollopre:
[...]

PR #738, opened 2026-06-02
fix: [737] preserve Consumer identity on network recovery

jollopre:
When recovering a consumer whose callable is a Bunny::Consumer instance, use basic_consume_with instead of basic_consume so that the original object reference—and any on_cancellation callbacks registered on it—are preserved after reconnection.


jollopre:
#737

PR #744, opened 2026-08-27
Push the gem from CI via trusted publishing

eglitobias:
Publishes the gem from CI on a version tag using [trusted publishing](https://guides.rubygems.org/trusted-publishing/), same as ruby-amqp/amq-protocol#90.

Two things are needed before the first release goes through:

* a GitHub environment named `release` (Settings > Environments). Without it the job fails with `environment not found` before any step runs
* a trusted publisher for the `bunny` gem on RubyGems.org pointing at this repository, workflow `release.yml`, environment `release`

This does not use `rubygems/release-gem`, unlike amq-protocol. That action runs `bundle exec rake release`, and Bundler's `version_tag` is hardcoded to `v#{version}`. Bunny tags releases as `3.2.0`, so `already_tagged?` would never match, and the task would create a second `v3.3.0` tag and try to push it from a detached checkout. Building and pushing the gem directly keeps the tag convention as it is. Recent RubyGems versions sign trusted publishing pushes on their own, so attestations work either way.

The tag filter accepts `3.3.0` and `3.3.0.pre`, and the job refuses to publish unless the tag matches `Bunny::VERSION`, so a mistyped tag is a failed run instead of a bad release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: Push gem
+
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+*"
+
+permissions:
+  contents: read
+
+jobs:
+  push:
+    name: Push Bunny to RubyGems.org
+    # Tags pushed in a fork must not attempt a release
+    if: github.repository == 'ruby-amqp/bunny'
+    runs-on: ubuntu-latest
+
+    concurrency:
+      group: push-gem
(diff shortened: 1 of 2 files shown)


michaelklishin:
The CI failures are unrelated.

michaelklishin:
I have created a `release` environment.

michaelklishin:
Trusted Publishing is set up on rubygems.org ✅

PR #745, opened 2026-08-30
connect_timeout for the TLS handshake

loadkpi:
Bunny applies several timeouts:
- `connect_timeout` to the TCP connection step
- `read_timeout` and `write_timeout` to the read and write operations

But it does not apply a timeout to the TLS handshake.

`Bunny::Transport#connect` calls `OpenSSL::SSL::SSLSocket#connect`. This call can block the thread for an unlimited time. This condition occurs when a peer accepts the TCP connection, but does not complete the TLS handshake. One example of such a peer is a load balancer in front of a RabbitMQ node that is out of operation.

We had more than one incident of this type after network problems. In each incident, Bunny stopped consumption of messages from RabbitMQ and did not continue. The connection did not recover, and no error was visible. A restart of the process was the only solution.

## Steps to Reproduce

1. Start a TCP server that accepts connections, but does not send data:
`nc -lk 15671`
2. Start a TLS connection to this server:
```ruby
conn = Bunny.new(host: "127.0.0.1", port: 15671, tls: true,
verify_peer: false, connect_timeout: 5)
conn.start
```
3. Wait for more than 5 seconds.

Expected behavior: Bunny::Session#start raises a timeout
Actual behaviour: Bunny::Session#start blocks for an unlimited time.

## Changes:

* Adds `Bunny::SSLSocket#connect_with_timeout`. The method does the TLS handshake with a time limit
* The TCP connection step already raises `Bunny::ClientTimeout` on a timeout. The handshake step now raises the same exception class
* Use `connect_timeout: 0` to disable the time limit. This setting gives the old behavior

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,17 @@
 ## Changes between Bunny 3.3.0 and 3.4.0 (in development)
 
-No changes yet.
+### TLS Handshake is Now Covered by the Connection Timeout
+
+The TLS handshake was the only connection establishment step not limited
+by any timeout. A peer that accepted a TCP connection but never completed
+the handshake (e.g. a load balancer in front of an unresponsive node)
+could block `Bunny::Session#start`, and with it connection recovery,
+indefinitely.
+
+The handshake now uses the same `connect_timeout` setting as the TCP
+connection step (default: 30 seconds) and raises a recoverable
+`Bunny::ClientTimeout` when the timeout is exceeded. Use
+`connect_timeout: 0` to opt out.
 
 
 ## Changes between Bunny 3.2.0 and 3.3.0 (Aug 29, 2026)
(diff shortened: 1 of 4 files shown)


eglitobias:
@loadkpi Thanks you!

I pushed an alternative on [ tls-handshake-timeout-io-timeout](https://github.com/ruby-amqp/bunny/pull/746). Would you mind looking at it, and trying it where you hit this originally? Since openssl 3.3, `SSLSocket#connect` honours `IO#timeout`, so no non-blocking loop is needed, at the price of `openssl >= 3.3` in the gemspec.

One thing that applies to your version too: ClientTimeout is not wrapped in TCPConnectionFailed, so Session#start never reaches the second endpoint.

michaelklishin:
Thank you, I have chosen #746.

PR #746, opened 2026-08-31
Apply connect_timeout to the TLS handshake, require Ruby 3.2

eglitobias:
Alternative to https://github.com/ruby-amqp/bunny/pull/745

### DIFF
--- a/bunny.gemspec
+++ b/bunny.gemspec
@@ -26,6 +26,8 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency "amq-protocol", "~> 2.9"
   s.add_runtime_dependency "logger", "~> 1", ">= 1.7"
+  # 3.3 is the first version whose SSLSocket#connect honours IO#timeout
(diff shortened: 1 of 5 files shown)


bunny.gemspec
```diff
@@ -26,6 +26,8 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency "amq-protocol", "~> 2.9"
   s.add_runtime_dependency "logger", "~> 1", ">= 1.7"
+  # 3.3 is the first version whose SSLSocket#connect honours IO#timeout
+  s.add_runtime_dependency "openssl", ">= 3.3"
```
eglitobias, line 30: The dependency is temporary: Ruby 3.4 ships openssl 3.3.0 as a default gem, so it can be dropped once required_ruby_version reaches 3.4 (Ruby 3.3 is EOL in March 2027).

michaelklishin (approved):

PR #750, opened 2026-09-01
Send the full client certificate chain

### LINKED ISSUE #749: mTLS client cert chain not sent when tls_cert is a multi-PEM bundle
## Question

Can this project provide a fix so that `tls_cert` sends the full certificate chain (leaf + intermediate CA), not only the first PEM?

We use mTLS from Fluentd (`fluent-plugin-rabbitmq` / `fluent-plugin-amqp`) to RabbitMQ. Certificates are issued by an Intermediate CA (cert-manager / Kubernetes TLS secret):

- `tls.crt` = leaf, then intermediate(s)
- `tls.key` = leaf private key
- `ca.crt` = Root CA only

The broker trusts the Root only. For the handshake to succeed, the client needs to present the Intermediate as well.

## What happens today

`tls_cert` is loaded here:

https://github.com/ruby-amqp/bunny/blob/2.14.4/lib/bunny/transport.rb

```ruby
ctx.cert = OpenSSL::X509::Certificate.new(@tls_certificate) if @tls_certificate
```

That keeps only the first certificate in the file. When we point tls_cert at a multi-PEM chain, mTLS to a Root-only broker fails (tlsv1 alert unknown ca).

Observed with Bunny 2.14.4. The Fluentd plugins expose tls_cert / tls_key / tls_ca_certificates and pass them to Bunny; they do not set a chain on the SSL context.

Request
Will you provide a fix for this? If work is already planned, can you point us to the issue or target release so we can follow it?

Thanks.

eglitobias:
`OpenSSL::X509::Certificate.new` keeps only the first PEM, so a leaf + intermediate bundle in `tls_cert` lost the intermediates and mTLS against a root-only broker failed with `tlsv1 alert unknown ca`. `Certificate.load` reads the whole bundle: leaf into `ctx.cert`, the rest into `ctx.extra_chain_cert`. Single-certificate files are unaffected.

Fixes #749

### DIFF
--- a/lib/bunny/transport.rb
+++ b/lib/bunny/transport.rb
@@ -497,7 +497,12 @@ module Bunny
     end
 
     def initialize_tls_context(ctx, opts = {})
-      ctx.cert       = OpenSSL::X509::Certificate.new(@tls_certificate) if @tls_certificate
+      if @tls_certificate
+        leaf, *chain = OpenSSL::X509::Certificate.load(@tls_certificate)
+
+        ctx.cert = leaf
+        ctx.extra_chain_cert = chain unless chain.empty?
+      end
       ctx.key        = OpenSSL::PKey.read(@tls_key) if @tls_key
       ctx.cert_store = if @tls_certificate_store
                          @tls_certificate_store
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,13 @@
 ## Changes between Bunny 3.3.0 and 3.4.0 (in development)
 
+### Client Certificate Chains
+
+`tls_cert` can now point at a bundle of the leaf certificate and its intermediate CAs.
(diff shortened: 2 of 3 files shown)


lib/bunny/transport.rb
```diff
@@ -497,7 +497,12 @@ def read_client_key!
     end
 
     def initialize_tls_context(ctx, opts = {})
-      ctx.cert       = OpenSSL::X509::Certificate.new(@tls_certificate) if @tls_certificate
+      if @tls_certificate
+        leaf, *chain = OpenSSL::X509::Certificate.load(@tls_certificate)
```
michaelklishin, line 501: Oh wow, so we've been relying on an implicit splat (collection destructuring in Erlang, Rust terms) the entire time.

michaelklishin (approved):

michaelklishin:
Thank you, @eglitobias.

